PR #1, opened 2018-10-05
Return after http.Error

metalmatze:
After writing with http.Error to the http.ResponseWriter we need to return.

/cc @s-urbaniak


s-urbaniak:
good catch :+1:

PR #4, opened 2018-11-15
README.md: Document risk outside scope of project

### LINKED ISSUE #2: Document security issue with honor_labels option
When Prometheus scrapes a target it applies labels to the metrics exposed by that target. To resolve the conflict between clashing label keys, Prometheus offers the [honor_labels](https://prometheus.io/docs/prometheus/latest/configuration/configuration/#scrape_config).

Let's say there is a multi-tenant Kubernetes cluster with a global Prometheus Operator and Prometheus. Prometheus is tenant aware via the prom-label-proxy. In addition let's assume that each tenant controls their Prometheus Operator ServiceMonitor.

In this scenario a malicious tenant A could set the `honor_labels` option to true and pollute the metrics of tenant B by applying the label `namespace=B` to each of their exposed metrics.

I don't see a way how the prom-label-proxy could prevent this issue. But I think documenting the concern is a good firth step. What are your thoughts?

mxinden:
Closes #2

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,4 +2,19 @@
 
 The prom-label-proxy enforces a given label in a given PromQL proxy.
 
-This proxy does not perform authentication or authorization, this has to happen before the request reaches this proxy.
+This proxy does not perform authentication or authorization, this has to happen
+before the request reaches this proxy. The
+[kube-rbac-proxy](https://github.com/brancz/kube-rbac-proxy) is an example for
+such an additional building block.
+
+
+Risks outside the scope of this project:
+
+- If a tenant controls its scrape target configuration the tenant can set
+  arbitrary labels via its relabelling configuration, thereby being able to
+  pollute other tenant's metrics.
+
+- If the ingestion configuration
+  [honor_labels](https://prometheus.io/docs/prometheus/latest/configuration/configuration/#scrape_config)
+  is set for a tenant's target, that target can pollute other tenant's metrics
+  as Prometheus respects any labels exposed by the target.


README.md
```diff
@@ -2,4 +2,14 @@
...
-This proxy does not perform authentication or authorization, this has to happen before the request reaches this proxy.
+This proxy does not perform authentication or authorization, this has to happen
+before the request reaches this proxy.
+
+Risks outside the scope of this project:
+
+- Tenants controlling ingestion options, e.g: If the system at hand
+  differentiates between two tennants (A and B), and tenant A has the ingestion
+  configuration
+  [honor_labels](https://prometheus.io/docs/prometheus/latest/configuration/configuration/#scrape_config)
```
brancz: This is true, but for the sake of completeness I would also mention, that if a user controls the configuration (such as relabelling) for a target the same risk exists.

mxinden: Adjusted beginning of list item and added relabelling notice to the end of the item. What do you think @brancz?

brancz: Sorry for being nit-picky, I think users don't grasp at first that those are distinct things:

* Controlling the scrape config allows people to set labels via relabelling therefore allowing them to pollute other tenants.
* If `honor_labels` is set, that allows targets to pollute the other tenants as Prometheus will respect any labels the target exposes.

README.md
```diff
@@ -2,4 +2,15 @@
 
 The prom-label-proxy enforces a given label in a given PromQL proxy.
 
-This proxy does not perform authentication or authorization, this has to happen before the request reaches this proxy.
+This proxy does not perform authentication or authorization, this has to happen
+before the request reaches this proxy.
```
metalmatze:
```suggestion
before the request reaches this proxy. The [kube-rbac-proxy](https://github.com/brancz/kube-rbac-proxy) is an example for doing this.
```

README.md
mxinden: No worries at all. I am very happy for the feedback!

mxinden:
@metalmatze @brancz updated, would you mind taking another look?

brancz:
perfect :+1: 

/lgtm

PR #16, opened 2019-12-19
Add support for /api/v1/rules

simonpasquier:
This change revisits heavily `injectproxy.routes` because prom-label-proxy needs to filter rules from the upstream's response rather than hacking/injecting query parameters in the incoming request. Hopefully the unit tests should make sure that there's no regression.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # prom-label-proxy
 
-The prom-label-proxy enforces a given label in a given PromQL proxy.
+The prom-label-proxy can enforce a given label in a given PromQL query or in Prometheus API responses.
 
 This proxy does not perform authentication or authorization, this has to happen before the request reaches this proxy. The [kube-rbac-proxy](https://github.com/brancz/kube-rbac-proxy) is an example for such an additional building block.
 
@@ -12,9 +12,17 @@ Risks outside the scope of this project:
 
 ## How does this project work?
 
-What this proxy does is it proxies the `/federate`, `/api/v1/query`, `/api/v1/query_range` prometheus endpoints and ensures that a particular label is enforced in the particular query.
+This application proxies the `/federate`, `/api/v1/query`, `/api/v1/query_range`, `/api/v1/rules` Prometheus endpoints and ensures that a particular label is enforced in the particular request and response.
 
-In the case of the federate endpoint, it ensures that all selectors passed as matchers to the federate endpoint _must_ contain that exact match of the particular label (and throws away all other matchers for the label). For the two query endpoints, it parses the PromQL expression and modifies all selectors in the same way. The label-key is configured as a flag on the binary and label-value is passed as a query parameter.
+Once again for clarity: this project only enforces a particular label in the respective calls to Prometheus, it in itself does not authenticate or authorize the requesting entity in any way, this has to be built around this project.
+
+### Federate endpoint
+
+The proxy ensures that all selectors passed as matchers to the `/federate` endpoint _must_ contain that exact match of the particular label (and throws away all other matchers for the label).
+
+### Query endpoints
+
+For the two query endpoints (`/api/v1/query` and `/api/v1/query_range`), the proxy parses the PromQL expression and modifies all selectors in the same way. The label-key is configured as a flag on the binary and the label-value is passed as a query parameter.
 
 For example, if requesting the PromQL query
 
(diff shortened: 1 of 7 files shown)


metalmatze:
Generally lgtm but I don't think I'm the audience to really review this new feature :)

README.md
```diff
@@ -12,9 +12,17 @@ Risks outside the scope of this project:
 
 ## How does this project work?
 
-What this proxy does is it proxies the `/federate`, `/api/v1/query`, `/api/v1/query_range` prometheus endpoints and ensures that a particular label is enforced in the particular query.
+What this proxy does is it proxies the `/federate`, `/api/v1/query`, `/api/v1/query_range`, `/api/v1/rules` Prometheus endpoints and ensures that a particular label is enforced in the particular request and response.
```
bwplotka:
```suggestion
This project is a proxy that proxies the `/federate`, `/api/v1/query`, `/api/v1/query_range`, `/api/v1/rules` Prometheus endpoints and ensures that a particular label is enforced in the certain request and response.
```

bwplotka: What about `/api/v1/alerts`?

injectproxy/rules.go
```diff
@@ -0,0 +1,192 @@
...
+	Value       string        `json:"value"`
+}
+
+func (r *routes) rules(resp *http.Response) error {
+	if resp.StatusCode != http.StatusOK {
+		// Pass non-200 responses as-is.
+		return nil
+	}
+
+	lvalue := labelValue(resp.Request.Context())
```
bwplotka: Here we check and panic, but in `query` forwarded we don't. Worth to be consistent? (:  Maybe `mustLabelValue` everywhere?

injectproxy/rules.go
```diff
@@ -0,0 +1,192 @@
...
+			filtered = append(filtered, rg)
+		}
+	}
+
+	if err := apir.setData(&rulesData{RuleGroups: filtered}); err != nil {
+		return errors.Wrap(err, "can't set rules data")
+	}
+
+	var buf bytes.Buffer
+	enc := json.NewEncoder(&buf)
```
bwplotka: nit: Can be inline, variable not reused.

injectproxy/rules.go
```diff
@@ -0,0 +1,192 @@
...
+	Warnings  []string        `json:"warnings,omitempty"`
+}
+
+func getAPIResponse(resp *http.Response) (*apiResponse, error) {
+	if resp.StatusCode != http.StatusOK {
+		return nil, fmt.Errorf("unexpected status code: %d", resp.StatusCode)
+	}
+
+	var apir apiResponse
+	dec := json.NewDecoder(resp.Body)
```
bwplotka: ditto

README.md
simonpasquier: for another PR

injectproxy/rules.go
simonpasquier: :+1:

injectproxy/routes.go
```diff
@@ -1,47 +1,93 @@
...
-func (r *routes) query(w http.ResponseWriter, req *http.Request) {
-	q := req.URL.Query()
-	labelValue := q.Get(r.label)
-	if labelValue == "" {
-		http.Error(w, fmt.Sprintf("Bad request. The %q query parameter must be provided.", r.label), http.StatusBadRequest)
-		return
+func (r *routes) ModifyResponse(resp *http.Response) error {
+	m, found := r.modifiers[resp.Request.URL.Path]
+	if !found {
+		return nil
```
s-urbaniak: we are not explicitly setting 404 here. Looking at https://golang.org/pkg/net/http/httputil/#ReverseProxy I don't see explicit behavior in case of a `nil` return. Can we reference/document what happens from here with the response?

injectproxy/routes_test.go
```diff
@@ -9,21 +10,62 @@ import (
...
-type mockHandler struct {
-	values url.Values
+// checkQueryParameterHandler verifies that the request contains the given parameter key/value.
+func checkQueryParameterHandler(key string, value string) http.HandlerFunc {
+	return http.HandlerFunc(func(w http.ResponseWriter, req *http.Request) {
+		values, err := url.ParseQuery(req.URL.RawQuery)
+		if err != nil {
+			http.Error(w, fmt.Sprintf("unexpected error: %v", err), http.StatusInternalServerError)
+		}
+		if len(values[key]) != 1 {
```
s-urbaniak: i guess if we don't return, `values` will be nil, no?

s-urbaniak: what i am trying to asl, if `err != nil`, do we still need to assert these invariants?

s-urbaniak:
/lgtm
this is looking great! :tada: i am just having two questions.

injectproxy/routes.go
simonpasquier: Sure i'll add this in a following PR. AFAIU when `ModifyResponse` doesn't return an error then the response is forwarded as-is to the client.

injectproxy/routes_test.go
simonpasquier: `values` is a `map[string][]string` so the test is mostly making sure that we don't get multiple values.

simonpasquier: But it should be more explicit for the reader :-)

PR #18, opened 2020-01-08
Add support for /api/v1/alerts

simonpasquier:


injectproxy/routes.go
```diff
@@ -59,6 +61,7 @@ func (r *routes) ServeHTTP(w http.ResponseWriter, req *http.Request) {
 func (r *routes) ModifyResponse(resp *http.Response) error {
 	m, found := r.modifiers[resp.Request.URL.Path]
 	if !found {
+		// Return the server's response unmodified.
```
s-urbaniak: :+1:

injectproxy/rules.go
```diff
@@ -185,3 +173,54 @@ func (r *routes) rules(resp *http.Response) error {
 
 	return nil
 }
+
+func (r *routes) rules(resp *http.Response) error {
+	return modifyAPIResponse(resp, func(b []byte) (interface{}, error) {
```
s-urbaniak: just a small nit: I really liked the simple idiom of the `rules(resp *http.Response)` modifier. Instead of converting a `http response` modifier here, we could modify `modifyAPIResponse` in a way that it converts a `func(*http.Response) error` into a `func(*apiResponse) error` .

Then we could have a nice and easy `func (r *routes) rules(resp *apiResponse) error` signature here without passing callbacks.

I have a proposal in a branch, let me know if this is a bit simpler, I will past a commit in a second.

s-urbaniak: @simonpasquier https://github.com/openshift/prom-label-proxy/commit/5809220166f1a0065e8716138b5ad61990ab9164

PR #20, opened 2020-01-15
Add support for /api/v2/silences

simonpasquier:
This change adds supports for Alertmanager silences. It enforces a given label for requests that list, create, update or create silences.


injectproxy/routes.go
```diff
@@ -41,21 +46,16 @@ func NewRoutes(upstream *url.URL, label string) *routes {
...
-		return
-	}
-
 	lvalue := req.URL.Query().Get(r.label)
 	if lvalue == "" {
 		http.Error(w, fmt.Sprintf("Bad request. The %q query parameter must be provided.", r.label), http.StatusBadRequest)
 		return
 	}
 	req = req.WithContext(withLabelValue(req.Context(), lvalue))
+	//FIXME(simonpasquier): this doesn't do anything, we need to update req.URL.RawQuery.
```
s-urbaniak: oh :-D

injectproxy/silences.go
```diff
@@ -0,0 +1,158 @@
...
+	"github.com/prometheus/alertmanager/pkg/labels"
+)
+
+func (r *routes) silences(w http.ResponseWriter, req *http.Request) {
+	switch req.Method {
+	case "GET":
+		r.listSilences(w, req)
+	case "POST":
+		r.postSilence(w, req)
+	}
```
s-urbaniak: can we quickly document here the expected behavior in case `req.Method != "GET" && req.Method != "POST"` ?

injectproxy/silences.go
```diff
@@ -0,0 +1,158 @@
...
+}
+
+func (r *routes) listSilences(w http.ResponseWriter, req *http.Request) {
+	var (
+		q        = req.URL.Query()
+		modified = []string{
+			(&labels.Matcher{Type: labels.MatchEqual, Name: r.label, Value: mustLabelValue(req.Context())}).String(),
```
s-urbaniak: nit: could we write this in multiple lines for readability?
```
&labels.Matcher{
  Type: labels.MatchEqual,
  Name: r.label,
  Value: mustLabelValue(req.Context()),
}.String(),
```

s-urbaniak:
@simonpasquier looking good so far, thanks a lot! do you mind to split the vendoring commit?

injectproxy/silences.go
simonpasquier: Agreed. Though it shouldn't happen because of the new HTTP routing logic (https://github.com/openshift/prom-label-proxy/pull/20/files#diff-70a97a8db748a07af11564c1ac089aabR36-R37), it's more clear to have it called out explicitly.

injectproxy/silences.go
simonpasquier: done

injectproxy/silences.go
s-urbaniak: I guess the edge case that we cover here now, is when the set of methods declared in `enforceMethods` does not match the set of handled methods here. Thanks for handling it :+1:

PR #28, opened 2020-05-16
Update deps to prometheus 2.18.1

vsliouniaev:
Working on upgrading coreos/prometheus-operator to support `absent_over_time` in admission webhooks to work.


s-urbaniak:
@vsliouniaev thanks a lot for the contribution! `go mod verify` fails unfortunately, do you mind to recheck?

vsliouniaev:
Thanks for the call-out, looks like everything is passing now.

lilic:
Sorry we are currently in a feature freeze cycle so this might need to wait until master branch opens.

/hold

go.mod
```diff
@@ -1,11 +1,14 @@
 module github.com/openshift/prom-label-proxy
 
-go 1.12
+go 1.13
+
+// v2.18.1
+replace github.com/prometheus/prometheus => github.com/prometheus/prometheus v0.0.0-20200507164740-ecee9c8abfd1
```
simonpasquier, line 6: I'd rather have `github.com/prometheus/prometheus v1.8.2-0.20200507164740-ecee9c8abfd1` in the `require` section and avoid using `replace`. WDYT?

vsliouniaev: Thanks for the pointer!

lilic:
@vsliouniaev do you mind updating prom-operator to use this instead of you fork, thanks!

PR #31, opened 2020-06-16
Finish migration to prometheus-community

lilic:
I was not sure what to do with the Dockerfile, we can sort that in a follow up PR, as currently the go.mod file is not allowing the correct new import path.

cc @brancz

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,4 +1,4 @@
-module github.com/openshift/prom-label-proxy
+module github.com/prometheus-community/prom-label-proxy
 
 go 1.13
 
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,7 @@
 PHONY: all
 all: check-license build generate test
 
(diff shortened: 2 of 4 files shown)


brancz:
Thanks!

go.mod
```diff
@@ -1,10 +1,11 @@
-module github.com/openshift/prom-label-proxy
+module github.com/prometheus-community/prom-label-proxy
 
 go 1.13
 
 require (
 	github.com/go-openapi/runtime v0.19.15
 	github.com/go-openapi/strfmt v0.19.5
+	github.com/openshift/prom-label-proxy v0.1.0 // indirect
```
simonpasquier, line 8: there are probably .go files still importing from this.

lilic: No think it was my local import that was wrong, should be fixed now, thanks!

lilic: PTAL :)

PR #39, opened 2020-09-02
injectproxy: allow multiple enforcement matchers

s-urbaniak:
Currently, the SetRecursive method does not work with multiple matchers.
This fixes it by introducing a small intermediate struct
and adds unit tests.

Signed-off-by: Sergiusz Urbaniak <sergiusz.urbaniak@gmail.com>


s-urbaniak:
I missed one edge case, looking into it.

s-urbaniak:
@simonpasquier edge case fixed, PTAL :-)

s-urbaniak:
cc'ing @lilic @brancz @paulfantom for a broader review, PTAL :-)

injectproxy/enforce.go
```diff
@@ -0,0 +1,131 @@
...
+	return &Enforcer{
+		labelMatchers: entries,
+	}
+}
+
+// EnforceNode walks the given node recursively
+// and enforces the given label enforcer on it.
+//
+// Whenever a parser.MatrixSelector or parser.VectorSelector AST node is found,
+// their label enforcer are being potentially modified.
```
squat, line 43: tiny grammar nit: we need correspondence between `are` and the singular `label enforcer`

injectproxy/enforce.go
```diff
@@ -0,0 +1,131 @@
...
+//
+// If a node label matcher equals the name with one of the given enforcer,
```
squat, line 45: This comment is a bit hard for me to understand. Can we rephrase? Maybe:
```
If a node's label matcher has the same name as a label matcher of the given enforcer, then it will be replaced.
```

injectproxy/enforce.go
```diff
@@ -0,0 +1,131 @@
...
+		labelMatchers: entries,
+	}
+}
+
+// EnforceNode walks the given node recursively
+// and enforces the given label enforcer on it.
+//
+// Whenever a parser.MatrixSelector or parser.VectorSelector AST node is found,
+// their label enforcer are being potentially modified.
+//
```
squat, line 44: Let's remove this newline, since the next sentence actually refines the sentence above.

injectproxy/enforce.go
squat: Is it the label enforcer of the node that is being modified? Or the label matcher of the node?

s-urbaniak: @squat ah, thank you for grammar corrections :pray: it is the latter, the label matcher of node. The enforcer is immutable it is "configuration data" if that makes sense.

injectproxy/enforce.go
s-urbaniak: sounds good :+1:

injectproxy/enforce.go
s-urbaniak: fixed

brancz (approved):
lgtm but giving others another chance to review

lilic (approved):
lgtm

Issue #41, opened 2020-10-04
Build/fix GitHub releases and Docker images


Tom-Fawcett:
Hey :slightly_smiling_face: 

I can't seem to find any recent GitHub releases or Docker images for this repo.

Looks like the CircleCI publish_master job has been [failing on the docker login](https://app.circleci.com/pipelines/github/prometheus-community/prom-label-proxy/24/workflows/4eec4bea-9f38-4f5e-9810-87df5af30d52/jobs/54/parallel-runs/0/steps/0-109) step.

I also noticed that [the Makefile mentions bind-exporter](https://github.com/prometheus-community/prom-label-proxy/blob/652a2d90c09f076448bd6777d17ae4f4c0f2dd05/Makefile#L22), which is presumably a mistake.

lilic:
Thanks for the issue!

@simonpasquier if I remember you finished up the release transfer, do you mind having a look at the failing docker login, thanks!

Tom-Fawcett:
Thanks for everyone's help with this :slightly_smiling_face:

Unfortunately it looks like CircleCI is still failing

PR #43, opened 2020-10-07
Fix container image name in Makefile

simonpasquier:
We accidentally pushed our image to https://quay.io/repository/prometheuscommunity/bind-exporter :disappointed:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -19,7 +19,7 @@ include Makefile.common
 
 STATICCHECK_IGNORE =
 
-DOCKER_IMAGE_NAME ?= bind-exporter
+DOCKER_IMAGE_NAME ?= prom-label-proxy
 
 .PHONY: run-curl-container
 run-curl-container:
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -19,7 +19,7 @@ include Makefile.common
 
 STATICCHECK_IGNORE =
 
-DOCKER_IMAGE_NAME ?= bind-exporter
```
lilic, line 22: Oops, but good catch!

lilic (approved):
lgtm

Makefile
simonpasquier: In fact it was mentioned by @Tom-Fawcett in https://github.com/prometheus-community/prom-label-proxy/issues/41 (thanks!) but I missed the info...

simonpasquier:
hmm, the push to quay.io has failed, probably an authentication problem. I'll look at it tomorrow presumably.

simonpasquier:
> hmm, the push to quay.io has failed, probably an authentication problem. I'll look at it tomorrow presumably.

The creds were indeed invalid. It should be fixed now but restarting the failed job doesn't work. It might be that the bad creds are still in the job's cache. Lets see if it works once #44 is merged.

PR #45, opened 2020-10-08
.circleci/config.yml: don't use org context

simonpasquier:
Environment variables defined in the organization's context take
[precedence](https://circleci.com/docs/2.0/env-vars/#order-of-precedence) over variables defined in the project. We use Quay
credentials (QUAY_LOGIN/QUAY_PASSWORD) that are scoped to the project
so we shouldn't pull the variables from the context.


lilic (approved):
lgtm

PR #48, opened 2020-10-21
Added Passthrough option for exposing upstream UI while ensuring multi-tenancy isolation.

bwplotka:
Also: Added example, and moved Frederic's example to kube-rbac-proxy dir.

This option is extremely useful for multi-tenancy without requiring Grafana, allowing to use Thanos, Prometheus UIs etc. Might be unsafe, if we miss certain APIs that's why passthrough is for non API endpoints and disabled by default.

Signed-off-by: Bartlomiej Plotka <bwplotka@gmail.com>


injectproxy/rules_test.go
```diff
@@ -658,7 +658,7 @@ func TestRules(t *testing.T) {
 		t.Run(fmt.Sprintf("%s=%s", proxyLabel, tc.labelv), func(t *testing.T) {
 			m := newMockUpstream(tc.upstream)
 			defer m.Close()
-			r := NewRoutes(m.url, proxyLabel)
+			r := NewRoutes(m.url, proxyLabel,false)
```
s-urbaniak, line 661: as per CI failure:
```
-			r := NewRoutes(m.url, proxyLabel,false)
+			r := NewRoutes(m.url, proxyLabel, false)
```

s-urbaniak:
I think this is a nice (although admittedly dangerous) feature! The only thing I am wondering though if this logic cannot be handled by nginx (or another proxy) natively, where `/api` requests would be dispatched to prom-label-proxy, and all other requests directly to the upstream target, be it Prometheus or Thanos, but having it in Go code is definitely more readable/maintainable.

I am curious about thoughts from others, cc @paulfantom @simonpasquier @lilic @brancz

bwplotka:
Yes, I thought about it, however, this would make it quite hard to use with nginx/caddy/envoy. Reasons:

* You are cannot isolate upstream from ingress, as some nginx has to access both prom-label-proxy and upstream for certain requests
* It's very special knowledge of what paths should be enforced, what is not. All is manually specified and very fragile. It would awesome to have only one well-maintained place like this, not many.

examples/caddy-port-based/Caddyfile
```diff
@@ -0,0 +1,13 @@
...
+reverse_proxy 127.0.0.1:8080
+
+}
+
+:8082 {
+
+rewrite * ?{query}&job=prometheus
+reverse_proxy 127.0.0.1:8080
+
+}
```
kakkoyun, line 13:
```suggestion
}

```

bwplotka:
Fixed (:

s-urbaniak:
From my side this LGTM :+1: As this contribution has bigger ramifications though I'd love to have another set of eyes from the other maintainers.

brancz:
I don't think this is safe to do, the targets page and configuration page may contain somewhat sensitive data.

bwplotka:
With other PR, those will return "not implemented" return code so that's ok
- until we implement tenancy for those.

Kind Regards,
Bartek Płotka (@bwplotka)

bwplotka:
Oh actually I read just now `the targets page and configuration page may contain somewhat sensitive data`. This makes sense.

I believe you are right. Full pass through is too dangerous. I would propose something else: path allowlist for passthrough with safe default of UI resources only. WDYT? 🤗 
If that makes sense, will update this PR.

Przemek, can you provide more info what exactly paths it would be necessary (and safe) to pass through in your case?

kprzemek:
@bwplotka I will try to explain it simply ;-)

I am doing a large PoC for one of the banks in my country.
In fact, they only need pasthrough for one purpose - setting the label on the namespace (isolation to a specific tenant in the query). Access to the Targets and Configuration page is absolutely not necessary.

An example from my Caddy (prom-label-proxy: 10.11.20.22:29090):
11.10.20.22:9092 {
    rewrite *? {query} & namespace = project1
    reverse_proxy 10.11.20.22:29090
}
11.10.20.22:9093 {
    rewrite *? {query} & namespace = project2
    reverse_proxy 10.11.20.22:29090
}

    Thanks to this, they can do DataSource in Grafana, for example: http://10.11.20.22:9092 - project1-DS or http://10.11.20.22:9093 - project2-DS. Isolation at the proxy level will ensure that in Grafana, a user with a specific DataSource [IP: Port for example: http://10.11.20.22:9092] will not be able to see someone else's (other namespace) data.

Everything else can go like through a revese proxy:
11/10/20.22:9091 {
    reverse_proxy 10/11/20.22:9090
}

In this case, it will be the DataSource address in Grafana for Administrators (for example: http://10.11.20.22:9091 - admin-DS) who have unlimited access to all namespaces / tenents.

Prom-label-proxy seems the most appropriate. By the way, it is connected to the Prometheus community. Everything under one roof ;-)

CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+## Unreleased
+
+* [FEATURE] Add -non-api-path-passthrough flag (disabled by default), which allows
+ exposing all non API resources from upstream without enforcing (e.g Prometheus UI).
```
brancz, line 4: Let's add a note how this is safe in regards to for example the targets and configuration pages.

brancz (commented):
Is the non-react UI entirely removed in Prometheus? Otherwise this will potentially leak information that was available through the legacy UI.

bwplotka:
> Is the non-react UI entirely removed in Prometheus?

It's not. But we could block access to that (or allow users to do so).

Still if we allowlist pass through for UI resources (JS, CSS) only (le'ts say) even if there is button for Target, Config etc - proxy will not route to that.

brancz:
Could we at least put the same not-implemented thing in there for the old UI as we have here for non explicit API endpoints, and add documentation that if people use this flag Prometheus has to be >= version x.y.z? Otherwise I feel that we will be creating a bunch of vulnerable Prometheus instances.

bwplotka:
Sure, I will propose something safe.

kprzemek:
So ok. 
When will the passthrough be in the main branch? ;-)
For our PoC, I need it in the official prom-label-proxy branch.

bwplotka:
Will try today / this week

bwplotka:
Sorry for delay, I addressed all comments.

cc @brancz @s-urbaniak @kprzemek PTAL 🤗

brancz:
lgtm

kakkoyun (approved):

kprzemek:
@bwplotka thank you for this commit.
I try tu use it, and I have some problem. Guys please help me. Maybe it's my browser's problem?
I try to open 3 paths: -unsafe-passthrough-paths="/graph,/stores,/new"
but when I try to open it in browser I have only one URL OK - /stores.
Sorry for the ignorance, but I'm stuck .
TGraph
TNewGraph
TStore

bwplotka:
It works, however you need to find out list of paths that you need to show. Check F12 console and network tab and see which ones are throwing 404 and pass them in this option if it's safe to do so from isolation point of view 🤗

kprzemek:
Works.
Thank you. It was /api/v1... I add them to -unsafe-passthrough-paths and voilà ;-)

PR #49, opened 2020-10-21
Added support for metadata APIs.

bwplotka:
Went ahead and implemented all APIs (: Disabled labels behind flag as there is only WIP PRs present right now.

Signed-off-by: Bartlomiej Plotka <bwplotka@gmail.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,7 +38,7 @@ go get github.com/prometheus-community/prom-label-proxy
 
 ## How does this project work?
 
-This application proxies the `/federate`, `/api/v1/query`, `/api/v1/query_range`, `/api/v1/rules`, `/api/v1/alerts` Prometheus endpoints as well as `/api/v2/silences` Alertmanager endpoint and it ensures that a particular label is enforced in the particular request and response.
+This application proxies the `/federate`, `/api/v1/query`, `/api/v1/query_range`, `/api/v1/series`, `/api/v1/labels`, `/api/v1/label/<name>/values`, `/api/v1/rules`, `/api/v1/alerts` Prometheus endpoints as well as `/api/v2/silences` Alertmanager endpoint and it ensures that a particular label is enforced in the particular request and response.
 
 Particularly, you can run `prom-label-proxy` with label `tenant` and point to example, demo Prometheus server e.g:
 
@@ -83,6 +83,13 @@ http_requests_total{namespace="b"}
 ```
 
 This is enforced for any case, whether a label matcher is specified in the original query or not.
+
+### Metadata endpoints
+
+Similar to query endpoint, for metadata endpoints `/api/v1/series`, `/api/v1/labels`, `/api/v1/label/<name>/values` the proxy injects the specified label all the provided `match[]` selectors. 
+
+NOTE: At the moment of creation `/api/v1/labels`, `/api/v1/label/<name>/values` does not support `match[]` so they are disabled by default. Use `-enable-label-apis` flag to enable 
+those (see https://github.com/prometheus/prometheus/issues/6178 for tracking development).
 
 ### Rules endpoint
 
--- a/injectproxy/routes.go
+++ b/injectproxy/routes.go
@@ -19,9 +19,15 @@ import (
 	"net/http"
 	"net/http/httputil"
 	"net/url"
+	"strings"
 
 	"github.com/prometheus/prometheus/pkg/labels"
 	"github.com/prometheus/prometheus/promql/parser"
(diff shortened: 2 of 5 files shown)


README.md
```diff
@@ -41,6 +41,13 @@ http_requests_total{namespace="b"}
 
 This is enforced for any case, whether a label matcher is specified in the original query or not.
 
+### Metadata endpoints
+
+Similar to query endpoint, for metadata endpoints `/api/v1/series`, `/api/v1/labels`, `/api/v1/label/<name>/values` the proxy inject the specified label in the `match` selector. 
+
+NOTE: At the moment of creation `/api/v1/labels`, `/api/v1/label/<name>/values` does not support `match[]` so they are disabled by default. Use `-enable-label-apis` flag to enable 
+those (see https://github.com/prometheus/prometheus/issues/6178 for tracking development).
```
s-urbaniak, line 49: i think this is a sane default (I was wondering why we don't enable it by default) :+1:

s-urbaniak:
LGTM :+1: Leaving final pass to another pair of eyes, cc @brancz @simonpasquier

README.md
bwplotka: Because it gives false impression that tenancy is ensured, but literally implementation of label matcher does not exists yet (WIP PRs only)

brancz:
How about we wait for the APIs to implement these and then bump the Prometheus version compatibility?

injectproxy/routes.go
```diff
@@ -41,11 +67,20 @@ func NewRoutes(upstream *url.URL, label string) *routes {
 		label:    label,
 	}
 	mux := http.NewServeMux()
-	mux.Handle("/federate", enforceMethods(r.federate, "GET"))
+	mux.Handle("/federate", enforceMethods(r.matcher, "GET"))
 	mux.Handle("/api/v1/query", enforceMethods(r.query, "GET", "POST"))
 	mux.Handle("/api/v1/query_range", enforceMethods(r.query, "GET", "POST"))
 	mux.Handle("/api/v1/alerts", enforceMethods(r.noop, "GET"))
 	mux.Handle("/api/v1/rules", enforceMethods(r.noop, "GET"))
+	mux.Handle("/api/v1/series", enforceMethods(r.matcher, "GET"))
```
simonpasquier, line 75: The `/federate` endpoint of the proxy throws away any existing `match[]` field in the incoming query and the outgoing query is just `match[]={<label>="<label value>"}`.

This doesn't really line up with the README.md that says:

> Similar to query endpoint, for metadata endpoints `/api/v1/series`, `/api/v1/labels`, `/api/v1/label/<name>/values` the proxy inject the specified label in the `match` selector.

bwplotka: Very good point, thanks for spotting. Let me change the PR to match docs and let's discuss. I think adding new `match[]` is what we want

bwplotka: actually ignored me, we need to either select within query or replace indeed

bwplotka:
Should be fixed :hugs:

bwplotka:
PTAL @s-urbaniak @simonpasquier

s-urbaniak:
LGTM :+1: Is there anything we need to do with respect to Frederic's comment in https://github.com/prometheus-community/prom-label-proxy/pull/49#issuecomment-716554688? I see both upstream issues are still open.

bwplotka:
> How about we wait for the APIs to implement these and then bump the Prometheus version compatibility?

I think this will be a bit cruel for users because:

* Most of the users rarely upgrade Prometheus and there is no need to block them, they can just disable those APIs
* We can leverage this on sooner

Being forward compatible is quite nice here I would say (:

brancz:
I'm ok with the flag, but I wouldn't merge and release this until the upstream API changes have been done.

yeya24:
Can we move forward with this pr if it is ready? Thanos has supported label matchers in https://github.com/thanos-io/thanos/commit/2508a70d8abf7b273e08060d9bdd56b7feef4a51 so it would be good to have this feature, too

bwplotka:
I think Prometheus is stil behind. Let's move Prometheus one forward as well (:

yeya24:
@bwplotka The prometheus pr is merged https://github.com/prometheus/prometheus/pull/8301

bwplotka:
I think so. Any objections to merge this? @s-urbaniak @brancz @lilic ?

brancz:
lgtm

brancz (approved):

PR #53, opened 2020-11-16
injectproxy: support queries via HTTP POST

squat:
Currently, the prom-label-proxy does not support proxying requests that
are sent via the HTTP POST method, whereas the Prometheus HTTP API
accepts query requests via both GET and POST.

This commit adds support for proxying queries via HTTP POST.
Note: these are the important cases to consider:
* POST requests can send the query via the URL query string;
* POST requests can send the query via the POST body;
* POST requests can send two queries, one in the POST body and another
in the URL query string; the Prometheus API's implementation implicitly
choses the query from the POST body, but relying on this implicit
behavior could lead to a potential vulnerability if the behavior ever
changes so the code today overrides queries in both locations
independetly; and
* POST requests may send some form values in the POST body, some in the
URL query string, and no query at all.

Signed-off-by: Lucas Servén Marín <lserven@gmail.com>

### DIFF
--- a/injectproxy/routes.go
+++ b/injectproxy/routes.go
@@ -16,6 +16,7 @@ package injectproxy
 import (
 	"context"
 	"fmt"
+	"io/ioutil"
 	"net/http"
 	"net/http/httputil"
 	"net/url"
@@ -159,27 +160,64 @@ func (r *routes) noop(w http.ResponseWriter, req *http.Request) {
 }
 
 func (r *routes) query(w http.ResponseWriter, req *http.Request) {
-	expr, err := parser.ParseExpr(req.FormValue(queryParam))
+	// The `query` can come in the URL query string and/or the POST body.
+	// For this reason, we need to try to enforcing in both places.
+	// Note: a POST request may include some values in the URL query string
+	// and others in the body. If both locations include a `query`, then
+	// enforce in both places.
+	q, f1, err := enforceValues(req.Context(), r.label, req.URL.Query())
 	if err != nil {
 		return
+	}
+	req.URL.RawQuery = q
+
+	var f2 bool
+	// Enforce the query in the POST body if needed.
+	if req.Method == http.MethodPost {
+		if err := req.ParseForm(); err != nil {
+			return
+		}
+		q, f2, err = enforceValues(req.Context(), r.label, req.PostForm)
+		if err != nil {
+			return
+		}
+		req.Body = ioutil.NopCloser(strings.NewReader(q))
+		req.ContentLength = int64(len(q))
+	}
+
+	// If no query was found, return early.
+	if !f1 && !f2 {
+		return
+	}
+
+	r.handler.ServeHTTP(w, req)
+}
+
+func enforceValues(ctx context.Context, name string, v url.Values) (string, bool, error) {
+	// If no values were given or no query is present,
+	// e.g. because the query came in the POST body
+	// but the URL query string was passed, then finish early.
+	if v.Get("query") == "" {
+		return v.Encode(), false, nil
+	}
+	expr, err := parser.ParseExpr(v.Get("query"))
+	if err != nil {
+		return "", true, err
 	}
 
 	e := NewEnforcer([]*labels.Matcher{
 		{
-			Name:  r.label,
+			Name:  name,
 			Type:  labels.MatchEqual,
-			Value: mustLabelValue(req.Context()),
+			Value: mustLabelValue(ctx),
 		},
 	}...)
 	if err := e.EnforceNode(expr); err != nil {
-		return
+		return "", true, err
 	}
 
-	q := req.URL.Query()
-	q.Set(queryParam, expr.String())
-	req.URL.RawQuery = q.Encode()
-
-	r.handler.ServeHTTP(w, req)
+	v.Set(queryParam, expr.String())
+	return v.Encode(), true, nil
 }
 
 // matcher ensures all the provided match[] if any has label injected. If none was provided, single matcher is injected.
--- a/injectproxy/routes.go
+++ b/injectproxy/routes.go
@@ -160,58 +160,59 @@ func (r *routes) noop(w http.ResponseWriter, req *http.Request) {
 }
 
 func (r *routes) query(w http.ResponseWriter, req *http.Request) {
+	e := NewEnforcer([]*labels.Matcher{{
+		Name:  r.label,
+		Type:  labels.MatchEqual,
+		Value: mustLabelValue(req.Context()),
+	}}...)
+
 	// The `query` can come in the URL query string and/or the POST body.
 	// For this reason, we need to try to enforcing in both places.
 	// Note: a POST request may include some values in the URL query string
 	// and others in the body. If both locations include a `query`, then
 	// enforce in both places.
-	q, f1, err := enforceValues(req.Context(), r.label, req.URL.Query())
+	q, found1, err := enforceQueryValues(e, req.URL.Query())
 	if err != nil {
 		return
 	}
 	req.URL.RawQuery = q
 
-	var f2 bool
+	var found2 bool
 	// Enforce the query in the POST body if needed.
 	if req.Method == http.MethodPost {
 		if err := req.ParseForm(); err != nil {
 			return
 		}
-		q, f2, err = enforceValues(req.Context(), r.label, req.PostForm)
+		q, found2, err = enforceQueryValues(e, req.PostForm)
 		if err != nil {
 			return
 		}
+		// We are replacing request body, close previous one (ParseForm ensures it is read fully and not nil).
+		_ = req.Body.Close()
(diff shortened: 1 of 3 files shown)


injectproxy/routes.go
```diff
@@ -118,27 +120,65 @@ func (r *routes) noop(w http.ResponseWriter, req *http.Request) {
...
 func (r *routes) query(w http.ResponseWriter, req *http.Request) {
-	expr, err := parser.ParseExpr(req.FormValue("query"))
+	// The `query` can come in the URL query string and/or the POST body.
+	// For this reason, we need to try to enforcing in both places.
+	// Note: a POST request may include some values in the URL query string
+	// and others in the body. If both locations include a `query`, then
+	// enforce in both places.
+	q, f1, err := enforceValues(req.Context(), r.label, req.URL.Query())
 	if err != nil {
 		return
```
bwplotka, line 130: all of those errors I think should be reporter somehow :thinking:

injectproxy/routes.go
```diff
@@ -118,27 +120,65 @@ func (r *routes) noop(w http.ResponseWriter, req *http.Request) {
 }
 
 func (r *routes) query(w http.ResponseWriter, req *http.Request) {
-	expr, err := parser.ParseExpr(req.FormValue("query"))
+	// The `query` can come in the URL query string and/or the POST body.
+	// For this reason, we need to try to enforcing in both places.
```
bwplotka, line 124:
```suggestion
	// , so enforce in label in both places.
```

injectproxy/routes.go
```diff
@@ -118,27 +120,65 @@ func (r *routes) noop(w http.ResponseWriter, req *http.Request) {
...
+	// Note: a POST request may include some values in the URL query string
```
bwplotka, line 125: That sounds wrong, shouldn't we just expect one of ways?

injectproxy/routes.go
```diff
@@ -118,27 +120,65 @@ func (r *routes) noop(w http.ResponseWriter, req *http.Request) {
...
+	// enforce in both places.
+	q, f1, err := enforceValues(req.Context(), r.label, req.URL.Query())
 	if err != nil {
 		return
 	}
+	req.URL.RawQuery = q
+
+	var f2 bool
+	// Enforce the query in the POST body if needed.
+	if req.Method == http.MethodPost {
```
bwplotka, line 136: I would always do that. This is because GET can have body as well (: Wonder if that was not a leak before this impl then (I think Prom takes form no matter what, also for GETs)

https://stackoverflow.com/questions/978061/http-get-with-request-body

bwplotka: Also I wonder if we can just remove the upper one and always parse Form. This is because Form takes care of query as well: 

```
// ParseForm populates r.Form and r.PostForm.
//
// For all requests, ParseForm parses the raw query from the URL and updates
// r.Form.
//
// For POST, PUT, and PATCH requests, it also reads the request body, parses it
// as a form and puts the results into both r.PostForm and r.Form. Request body
// parameters take precedence over URL query string values in r.Form.
//
// If the request Body's size has not already been limited by MaxBytesReader,
// the size is capped at 10MB.
//
// For other HTTP methods, or when the Content-Type is not
// application/x-www-form-urlencoded, the request Body is not read, and
// r.PostForm is initialized to a non-nil, empty value.
//
// ParseMultipartForm calls ParseForm automatically.
// ParseForm is idempotent.
```

injectproxy/routes.go
```diff
@@ -118,27 +120,65 @@ func (r *routes) noop(w http.ResponseWriter, req *http.Request) {
...
+		if err := req.ParseForm(); err != nil {
+			return
+		}
+		q, f2, err = enforceValues(req.Context(), r.label, req.PostForm)
+		if err != nil {
+			return
+		}
+		req.Body = ioutil.NopCloser(strings.NewReader(q))
```
bwplotka, line 144: We should close req.Body first (no one does it for us).

bwplotka (commented):
Amazing, generally LGTM. 

Some nits though.

kakkoyun (approved):
💯 LGTM. Great PR.

injectproxy/routes.go
metalmatze: I think historically we once did but then someone used did it that other way and it broke. We definitely need to support both. :man_shrugging:

kakkoyun:
Apparently not that great 😅 @bwplotka and his laser eyes.

injectproxy/routes.go
squat: We don't do this before. I would prefer to introduce this in a followup

injectproxy/routes.go
squat: `Enforce in label` is not syntactically correct.i would prefer as is

injectproxy/routes.go
squat: This is literally how it works today in Prometheus because Prometheus uses r.FormValue

injectproxy/routes.go
squat: No we don't want to remove this. Again, values can come from EITHER one of the places OR both. There are some really funky things that can happen if someone puts a query into BOTH the query string AND the form body. Imagine that tomorrow Prometheus stops using FormValue and only reads the query from one of them, then we would only override the form body but keep the unescaped query in the query string. This would very easily lead to a vulnerability where the unescaped query in the URL is executed. Or imagine that a proxy in the middle of the promlabelproxy and Prometheus only uses the url query string. Same thing would happen. We need to be defensive and escape both queries. We cannot do this if we use FormValue, so we need to explicitly look for queries in both places

injectproxy/routes.go
bwplotka: sgtm

injectproxy/routes.go
bwplotka: "try" is also weird though. We don't try. We do or fail.

injectproxy/routes.go
bwplotka: Not really:

> Request body parameters take precedence over URL query string values in r.Form.

So there is no point in doing URL injection if there is form, we can just delete URL one if we want to be sure nothing leaks.

injectproxy/routes.go
bwplotka: > This would very easily lead to a vulnerability where the unescaped query in the URL is executed

Agree. This is why vote for clearing out query param completely.

injectproxy/routes.go
squat: Why? This is making assumptions about what the user wants. The user meant to send a query parameter? Great then escape that. The user also sent a form parameter? Great, then escape that. Otherwise you are mixing the semantics of the implementation detail of Go-s form parsing with business-logic just because both this proxy and Prometheus happen to be both written in Go. Escaping both ensures nothing leaks and that we don't assume one over the other.

injectproxy/routes.go
squat: Tested this thoroughly and it is not a vulnerability today due to shear luck because of the form parsing logic in go.

> but rather vote for clearing out query param completely on POST.

Why? The user sent both so let's escape both. We are dropping user data otherwise. Essentially are codifying the implementation logic of Go's form parsing as business logic of this proxy. They are two different domains and should not be mixed. If Go preferred query parameters over form parameters what would we do? And what if we used Python instead? etc

bwplotka:
I feel bad for blocking this and not reviewing for long time, so wanted to help @squat  (: 

Rebased for now (there were conflicts). Can fix last tiny nitty issues. (Want to move Obs API forward)

injectproxy/routes.go
bwplotka: Not a biggy.

injectproxy/routes.go
bwplotka: I think that a little bit more complex then it needs to be, but as long as safe - ok (:

injectproxy/routes.go
bwplotka: > Tested this thoroughly and it is not a vulnerability today due to shear luck because of the form parsing logic in go.

Thanks!

Added test case as well in https://github.com/prometheus-community/prom-label-proxy/pull/53/commits/c9da73397b6c60a3326bd6cf06b99261dadc27e5

injectproxy/routes.go
bwplotka: Addressed https://github.com/prometheus-community/prom-label-proxy/pull/53/commits/c9da73397b6c60a3326bd6cf06b99261dadc27e5 let me know @squat if that makes sense to you

bwplotka (approved):
Addressed literally 2 things in https://github.com/prometheus-community/prom-label-proxy/pull/53/commits/c9da73397b6c60a3326bd6cf06b99261dadc27e5 and cleaned `enforceValue` a bit, hope that makes sense

Let me know if that's ok to merge. LGTM to me (:

simonpasquier (approved):
:+1:

PR #54, opened 2020-12-15
Add option to get label value from an HTTP header

scoof:
This PR allows the client to send the label value as an HTTP header. Doing it this way, combined with #53 will allow this to be used for alerting in Grafana.

### DIFF
--- a/injectproxy/routes_test.go
+++ b/injectproxy/routes_test.go
@@ -108,6 +108,7 @@ func (m *mockUpstream) Close() {
 }
 
 const proxyLabel = "namespace"
+const proxyLabelParam = "ns"
 const proxyHeader = "X-Scope-OrgId"
 
 func TestEndpointNotImplemented(t *testing.T) {
@@ -115,10 +116,10 @@ func TestEndpointNotImplemented(t *testing.T) {
 		w.Write(okResponse)
 	}))
 	defer m.Close()
-	r := NewRoutes(m.url, proxyLabel, proxyHeader)
+	r := NewRoutes(m.url, proxyLabel, proxyLabelParam, proxyHeader)
 
 	w := httptest.NewRecorder()
-	req := httptest.NewRequest("GET", "http://prometheus.example.com/graph?namespace=ns1", nil)
+	req := httptest.NewRequest("GET", "http://prometheus.example.com/graph?ns=ns1", nil)
 
 	r.ServeHTTP(w, req)
 	resp := w.Result()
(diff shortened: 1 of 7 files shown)


simonpasquier:
:wave: could you rebase?

scoof:
Rebased

injectproxy/routes_test.go
```diff
@@ -421,3 +422,90 @@ func TestQuery(t *testing.T) {
...
+		expBody      []byte
+	}{
+		{
+			// No "namespace" parameter returns an error.
+			expCode: http.StatusBadRequest,
+		},
+		{
+			// Query with only query
+			labelv:       "default",
+			headerv:      "defaulth",
```
s-urbaniak, line 443: nit: i think for this test case this should be `headerv:      "",`

s-urbaniak:
@scoof thank you for the contribution! i just have one nit regarding a testcase, else lgtm.

injectproxy/routes_test.go
scoof: You're absolutely correct....

simonpasquier (commented):
Do we want to allow passing the label name only by the header? Currently you would have to set the `--label` flag to a non-empty value which is a bit confusing.

scoof:
If we want to do that, then I think we should have a separate flag to set the prometheus label, different from the query param. That would break backwards compatibility. Thoughts?

simonpasquier:
Personally I wouldn't be too worried about backwards compatibility as long as it's reflected in the changelog. I would expect that one should be able to choose between 1) read from query parameters only 2) read from HTTP header only or 3) read from query parameters then HTTP header if empty.

@s-urbaniak @brancz thoughts?

s-urbaniak:
hm, I think you're right, I was somehow assuming that `r.label` is just being used here, but it is all over the place. Yeah I agree, potentially adding another option just for specifying the query parameter sounds good to me.

scoof:
So --label --query-param and --header.
Would that make sense?
Then --label is required and at least one of --query-param and --header

scoof:
Still need to fix docs

scoof:
I haven't forgotten about this, but I'm quite busy with a new baby.

horazont:
Hey @scoof. I took the liberty to take this off your mental load and rebase it into current master in #64. If you find a minute, I’d be glad if you could confirm that you’re attributed appropriately.

scoof:
@horazont looks fine, thanks!

PR #64, opened 2021-04-27
Add option to get label value from an HTTP header

horazont:
This allows to use the proxy in settings where query parameters
cannot be used; this could be behind specific reverse proxies, or
when HTTP POST is required (see #53).

This work is heavily based on the work done by
Andreas Plesner <apj@mutt.dk> in [1]. I (Jonas Schäfer) ported
this to the current head of master.

   [1]: https://github.com/prometheus-community/prom-label-proxy/pull/54

This is a rebase of #54. Please review carefully. I only ran the test suite so far.

### DIFF
--- a/main.go
+++ b/main.go
@@ -32,14 +32,19 @@ func main() {
 		insecureListenAddress  string
 		upstream               string
 		label                  string
+		headerName             string
+		queryParam             string
 		enableLabelAPIs        bool
 		unsafePassthroughPaths string // Comma-delimited string.
 		errorOnReplace         bool
+		header                 string
 	)
 
 	flagset := flag.NewFlagSet(os.Args[0], flag.ExitOnError)
 	flagset.StringVar(&insecureListenAddress, "insecure-listen-address", "", "The address the prom-label-proxy HTTP server should listen on.")
 	flagset.StringVar(&upstream, "upstream", "", "The upstream URL to proxy to.")
+	flagset.StringVar(&queryParam, "query-param", "", "The query parameter to obtain the label value from. This or -header is required.")
+	flagset.StringVar(&headerName, "header", "", "The HTTP header name to obtain the label value from. This or -query-param is required.")
 	flagset.StringVar(&label, "label", "", "The label to enforce in all proxied PromQL queries. "+
 		"This label will be also required as the URL parameter to get the value to be injected. For example: -label=tenant will"+
 		" make it required for this proxy to have URL in form of: <URL>?tenant=abc&other_params...")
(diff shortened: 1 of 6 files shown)


hanjm:
It will be userful in grafana datasource, could any one can update for this? :)

horazont:
@hanjm It indeed is very useful with Grafana :).

I rebased onto current master.

kfox1111:
Any update on this?

main.go
```diff
@@ -51,12 +56,25 @@ func main() {
 		"API like targets or configuration. NOTE: \"all\" matching paths like \"/\" or \"\" and regex are not allowed.")
 	flagset.BoolVar(&errorOnReplace, "error-on-replace", false, "When specified, the proxy will return HTTP status code 400 if the query already contains a label matcher that differs from the one the proxy would inject.")
 
+	flagset.StringVar(&header, "header", "", "(Optional) An HTTP header to get the label value from.")
```
d9705996, line 59: The flag name header is duplicated and causes a panic

```go
panic: ./prom-label-proxy flag redefined: header

goroutine 1 [running]:
flag.(*FlagSet).Var(0xc00031e420, {0x9a9288, 0xc000313670}, {0x8fec55, 0x6}, {0x91a2ac, 0x36})
        flag/flag.go:879 +0x2f4
flag.(*FlagSet).StringVar(...)
        flag/flag.go:762
main.main()
        main.go:59 +0x4d9
```

m-yosefpor:
Hi @horazont . Are you still working on this?

main.go
horazont: Sorry, fixed. I only use this as a library, so I didn't notice.

horazont:
@m-yosefpor Thanks for the ping. I rebased onto current master, but I don't currently have time to test the rebase thoroughly.

main.go
```diff
@@ -32,14 +32,19 @@ func main() {
 		insecureListenAddress  string
 		upstream               string
 		label                  string
+		headerName             string
```
Wouter0100, line 35: There's a mix of `headerName` and `header` in this file. `headerName` is filled with the flag, but `header` is validated and actually used which - obviously - is always empty.

Wouter0100 (changes requested):

main.go
ThisIsQasim: You are right I had to use the following patch

```
diff --git a/main.go b/main.go
index d7b29dc..3628eaa 100644
--- a/main.go
+++ b/main.go
@@ -32,7 +32,6 @@ func main() {
                insecureListenAddress  string
                upstream               string
                label                  string
-               headerName             string
                queryParam             string
                enableLabelAPIs        bool
                unsafePassthroughPaths string // Comma-delimited stri
@@ -44,7 +43,7 @@ func main() {
        flagset.StringVar(&insecureListenAddress, "insecure-listen-ad
        flagset.StringVar(&upstream, "upstream", "", "The upstream UR
        flagset.StringVar(&queryParam, "query-param", "", "The query 
-       flagset.StringVar(&headerName, "header", "", "The HTTP header
+       flagset.StringVar(&header, "header", "", "The HTTP header nam
        flagset.StringVar(&label, "label", "", "The label to enforce 
                "This label will be also required as the URL paramete
                " make it required for this proxy to have URL in form
```

kfox1111:
Any movement on this patch?

PR #66, opened 2021-05-15
Reuse injectproxy package

Abhishek357:
Signed-off-by: Abhishek357 <abhisheksinghchauhan357@gmail.com>

We would like to reuse the injectproxy package in thanos. Related to [PR](https://github.com/thanos-io/thanos/pull/4141)

### DIFF
--- a/injectproxy/enforce.go
+++ b/injectproxy/enforce.go
@@ -21,7 +21,7 @@ import (
 )
 
 type Enforcer struct {
-	LabelMatchers map[string]*labels.Matcher
+	labelMatchers map[string]*labels.Matcher
 }
 
 func NewEnforcer(ms ...*labels.Matcher) *Enforcer {
@@ -32,7 +32,7 @@ func NewEnforcer(ms ...*labels.Matcher) *Enforcer {
 	}
 
 	return &Enforcer{
-		LabelMatchers: entries,
+		labelMatchers: entries,
 	}
 }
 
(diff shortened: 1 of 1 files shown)


injectproxy/enforce.go
```diff
@@ -111,18 +111,18 @@ func (ms Enforcer) EnforceNode(node parser.Node) error {
 	return nil
 }
 
-func (ms Enforcer) enforceMatchers(targets []*labels.Matcher) []*labels.Matcher {
```
bwplotka, line 114: Do you mind adding quick commentary what this public method does (Public methods should be commented always with comment started as `// EnforceMatchers ...`

injectproxy/enforce.go
```diff
@@ -21,7 +21,7 @@ import (
 )
 
 type Enforcer struct {
-	labelMatchers map[string]*labels.Matcher
```
bwplotka, line 24: Why we need access to this? It looks like code smell, as it leaks the enforcer internal logic. Aren't its methods enough? Are you missing some functionality that could be added as method?

bwplotka (commented):
Amazing, small nits, otherwise LGTM! (:

injectproxy/enforce.go
Abhishek357: We need to access this because enforceMatchers is a method of enforcer and to access it first we need to create an enforcer(in our use case in Thanos this enforcer contains the information about the tenant label name we want to replace and tenant access a user has as specified in the header)

injectproxy/enforce.go
Abhishek357: sure!

injectproxy/enforce.go
bwplotka: yea, but this method has access to this because it's a method on `Enforcer` itself, no?

Abhishek357: hmm, I think we can use [NewEnforcer](https://github.com/prometheus-community/prom-label-proxy/blob/38b1382d7368e994e39c0cb853f3f3c9300f5045/injectproxy/enforce.go#L27) to create a new enforcer.
Let me try it.

bwplotka: Well. I think we even have to! That's why we do constructors in Go (:

Abhishek357: ahh just refreshed those concepts thanks! definitely make sense.

injectproxy/enforce.go
```diff
@@ -111,7 +111,9 @@ func (ms Enforcer) EnforceNode(node parser.Node) error {
 	return nil
 }
 
-func (ms Enforcer) enforceMatchers(targets []*labels.Matcher) []*labels.Matcher {
+// EnforceMatchers replaces similar matcher present in
```
bwplotka, line 114: I would be specific what "similar" means, but it's good enough (:

bwplotka (approved):
LGTM, thanks!

injectproxy/enforce.go
Abhishek357: will keep that in mind.

PR #67, opened 2021-05-18
return error if existing label matcher in query would change

jan--f:
Before this commit, if the inject proxy is configured to replace label
foo and a query is encountered that has one or more label matchers for
the foo label, these matchers are silently replaced. E.g. a query like
`up{foo="default"}` would silently be altered to
`up{foo="replacement"}` and the query result returned.
With this change we return 400 instead.

Signed-off-by: Jan Fajerski <jfajersk@redhat.com>

### DIFF
--- a/injectproxy/enforce.go
+++ b/injectproxy/enforce.go
@@ -39,7 +39,7 @@ func NewEnforcer(errorOnReplace bool, ms ...*labels.Matcher) *Enforcer {
 	}
 }
 
-var ErrIllegalLabelMatcher = errors.New("Label matcher value conflicts with injected value")
+var ErrIllegalLabelMatcher = errors.New("label matcher value conflicts with injected value")
 
 // EnforceNode walks the given node recursively
 // and enforces the given label enforcer on it.
--- a/injectproxy/enforce.go
+++ b/injectproxy/enforce.go
@@ -14,6 +14,7 @@
 package injectproxy
 
 import (
+	"errors"
 	"fmt"
 
 	"github.com/prometheus/prometheus/pkg/labels"
@@ -21,10 +22,11 @@ import (
 )
 
 type Enforcer struct {
-	labelMatchers map[string]*labels.Matcher
+	labelMatchers  map[string]*labels.Matcher
+	errorOnReplace bool
 }
 
-func NewEnforcer(ms ...*labels.Matcher) *Enforcer {
+func NewEnforcer(errorOnReplace bool, ms ...*labels.Matcher) *Enforcer {
 	entries := make(map[string]*labels.Matcher)
 
 	for _, matcher := range ms {
@@ -32,9 +34,12 @@ func NewEnforcer(ms ...*labels.Matcher) *Enforcer {
 	}
 
 	return &Enforcer{
-		labelMatchers: entries,
+		labelMatchers:  entries,
+		errorOnReplace: errorOnReplace,
 	}
 }
+
+var ErrIllegalLabelMatcher = errors.New("Label matcher value conflicts with injected value")
 
 // EnforceNode walks the given node recursively
 // and enforces the given label enforcer on it.
@@ -97,12 +102,20 @@ func (ms Enforcer) EnforceNode(node parser.Node) error {
 	case *parser.MatrixSelector:
 		// inject labelselector
 		if vs, ok := n.VectorSelector.(*parser.VectorSelector); ok {
-			vs.LabelMatchers = ms.EnforceMatchers(vs.LabelMatchers)
+			var err error
+			vs.LabelMatchers, err = ms.EnforceMatchers(vs.LabelMatchers)
+			if err != nil {
+				return err
+			}
 		}
 
 	case *parser.VectorSelector:
 		// inject labelselector
-		n.LabelMatchers = ms.EnforceMatchers(n.LabelMatchers)
+		var err error
+		n.LabelMatchers, err = ms.EnforceMatchers(n.LabelMatchers)
+		if err != nil {
+			return err
+		}
 
 	default:
 		panic(fmt.Errorf("parser.Walk: unhandled node type %T", n))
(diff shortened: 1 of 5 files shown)


jan--f:
cc @simonpasquier, @s-urbaniak

simonpasquier:
> maybe its less disruptive if protected by a cli flag

it might be a good option indeed. There might be users who are fine with the current behavior which overrides the label matcher silently.

>  Currently this compares label matcher values as strings. So it doesn't take the matcher type into consideration or attempts to do anything when a regular expression is passed as the matcher value. Is this valid or should the proxy do more here?

I'm not sure, maybe we should reject negative matchers. If the provided query is `up{namespace!="foo"}`, it would turn into `up{namespace!="foo",namespace="foo"}` which would return no result and be surprising.

jan--f:
> I'm not sure, maybe we should reject negative matchers. If the provided query is `up{namespace!="foo"}`, it would turn into `up{namespace!="foo",namespace="foo"}` which would return no result and be surprising.

Luckily that's not quite the case. We actually parse the query and currently replace label matchers based on the label name, so `up{namespace!="foo"}` would become `up{namespace="foo"}`. However you still make a good point, just comparing the value isn't enough, we need to compare the match type as well.

s-urbaniak:
Generally I agree to this very useful feature :+1: 

> maybe its less disruptive if protected by a cli flag

Agreed, and we have to make sure that the default does not change behavior of existing prom-label-proxy deployments.

jan--f:
ping @simonpasquier @s-urbaniak kindly requesting another review.

injectproxy/routes.go
```diff
@@ -264,6 +275,9 @@ func (r *routes) query(w http.ResponseWriter, req *http.Request) {
 	// enforce in both places.
 	q, found1, err := enforceQueryValues(e, req.URL.Query())
 	if err != nil {
+		if err == ErrIllegalLabelMatcher {
+			http.Error(w, err.Error(), http.StatusForbidden)
```
squat, line 279: I wonder if this should rather be `StatusBadRequest`

injectproxy/routes.go
```diff
@@ -276,6 +290,9 @@ func (r *routes) query(w http.ResponseWriter, req *http.Request) {
 		}
 		q, found2, err = enforceQueryValues(e, req.PostForm)
 		if err != nil {
+			if err == ErrIllegalLabelMatcher {
+				http.Error(w, err.Error(), http.StatusForbidden)
```
squat, line 294: same here

main.go
```diff
@@ -48,6 +49,7 @@ func main() {
 	flagset.StringVar(&unsafePassthroughPaths, "unsafe-passthrough-paths", "", "Comma delimited allow list of exact HTTP path segments should be allowed to hit upstream URL without any enforcement."+
 		"This option is checked after Prometheus APIs, you can cannot override enforced API to be not enforced with this option. Use carefully as it can easily cause a data leak if the provided path is an important"+
 		"API like targets or configuration. NOTE: \"all\" matching paths like \"/\" or \"\" and regex are not allowed.")
+	flagset.BoolVar(&errorOnReplace, "error-on-replace", false, "When specified proxy will return 403 if the query already contains a label matcher that differs from the on the proxy would inject.")
```
squat, line 52:
```suggestion
	flagset.BoolVar(&errorOnReplace, "error-on-replace", false, "When specified proxy will return 403 if the query already contains a label matcher that differs from the one the proxy would inject.")
```

squat, line 52: nit: I think we can make this a full sentence:
```suggestion
	flagset.BoolVar(&errorOnReplace, "error-on-replace", false, "When specified, the proxy will return HTTP status code 403 if the query already contains a label matcher that differs from the on the proxy would inject.")
```

injectproxy/enforce.go
```diff
@@ -14,28 +14,33 @@
...
 	}
 
 	return &Enforcer{
-		labelMatchers: entries,
+		labelMatchers:  entries,
+		errorOnReplace: errorOnReplace,
 	}
 }
 
+var ErrIllegalLabelMatcher = errors.New("Label match value conflicts with injected value")
```
squat, line 42: In Golang we normally make errors start with a lower-case letter.
Also, for consistency this should be `matcher` I think 
```suggestion
var ErrIllegalLabelMatcher = errors.New("label matcher value conflicts with injected value")
```

squat (commented):
looks really clean to me overall :)

injectproxy/routes.go
jan--f: Yeah fair point. Since we don't actually check if the user is authorized or not `StatusBadRequest` seems more appropriate.

injectproxy/enforce.go
```diff
@@ -14,28 +14,33 @@
...
 	}
 
 	return &Enforcer{
-		labelMatchers: entries,
+		labelMatchers:  entries,
+		errorOnReplace: errorOnReplace,
 	}
 }
 
+var ErrIllegalLabelMatcher = errors.New("Label matcher value conflicts with injected value")
```
squat, line 42: Raising this again
```suggestion
var ErrIllegalLabelMatcher = errors.New("label matcher value conflicts with injected value")
```

jan--f: Sorry missed that one.

squat (approved):
I like it :) good work

injectproxy/enforce.go
```diff
@@ -111,13 +124,18 @@ func (ms Enforcer) EnforceNode(node parser.Node) error {
...
+// different (either by match type or value) an error is returned.
+func (ms Enforcer) EnforceMatchers(targets []*labels.Matcher) ([]*labels.Matcher, error) {
 	var res []*labels.Matcher
 
 	for _, target := range targets {
-		if _, ok := ms.labelMatchers[target.Name]; ok {
+		if matcher, ok := ms.labelMatchers[target.Name]; ok {
+			// matcher.String() returns something like "labelfoo=value"
+			if ms.errorOnReplace && matcher.String() != target.String() {
+				return res, ErrIllegalLabelMatcher
```
simonpasquier, line 137: do we want to tell which label is conflicting?

injectproxy/enforce.go
```diff
@@ -111,13 +124,18 @@ func (ms Enforcer) EnforceNode(node parser.Node) error {
 	return nil
 }
 
-// EnforceMatchers replaces similar matcher present in
-// target label matchers or appends if not present.
-func (ms Enforcer) EnforceMatchers(targets []*labels.Matcher) []*labels.Matcher {
+// EnforceMatchers appends the configured label matcher if not present.
+// If the label matcher that is to be injected is present (by labelname) but
+// different (either by match type or value) an error is returned.
```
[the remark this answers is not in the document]
jan--f: True, added that detail.

injectproxy/enforce.go
jan--f: Yes, good point. Added a commit to that effect.

simonpasquier (approved):
lgtm

PR #71, opened 2021-06-22
go,vendor: bump Prometheus dependency

squat:
This commit bumps the dependency on Prometheus to the commit from the
v2.28.0 tag
(https://github.com/prometheus/prometheus/releases/tag/v2.28.0). The
reason for bumping this is to take advantage of the updates to the
PromQL parser. Without this update, the prom-label-proxy cannot enforce
labels on queries that use any new features, e.g. queries that include:
* the last_over_time, sgn, clamp functions, added in
  https://github.com/prometheus/prometheus/pull/8457;
* the @ operator, added in
  https://github.com/prometheus/prometheus/pull/8121; or
* the group aggregation operator, added in
  https://github.com/prometheus/prometheus/pull/7480.

Currently, queries including these functions produce errors like:
`parse error: unknown function with name "group"`

Signed-off-by: Lucas Servén Marín <lserven@gmail.com>


squat:
Failure is due to: https://github.com/prometheus/alertmanager/issues/2630
Once we figure this issue out, we can decide how to fix the test case here.

bwplotka:
Nice catch and LGTM with the fix.

We can import this or we can wait for merge of fix and get main: https://github.com/prometheus/alertmanager/issues/2630

bwplotka:
Blocker merged

bwplotka (approved):

PR #77, opened 2021-08-25
README: add authzed example

jzelinskie:
Hey there 👋 

First of all a big thanks to the community for building and maintaining an incredibly useful project.

[We built our own proxy](https://github.com/authzed/prom-authzed-proxy) based on prom-label-proxy that enables our users to query our internal Prometheus stack securely.
I think it'd be useful mention this in the README, because it uses this project as library, rather running it standalone and putting another proxy in front of it.

I'm also open to discussing where we found it difficult/annoying to use this project as a library and how we could improve that experience, if it's something the community is interested in.

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,7 +5,7 @@
 The prom-label-proxy can enforce a given label in a given PromQL query, in Prometheus API responses or in Alertmanager API requests. As an example (but not only),
 this allows read multi-tenancy for projects like Prometheus, Alertmanager or Thanos.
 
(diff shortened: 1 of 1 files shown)


jzelinskie:
Hey, not sure if I need to do anything to unblock this.

The biggest annoyance was simply the lack of access to [the ReverseProxy](https://github.com/prometheus-community/prom-label-proxy/blob/v0.4.0/injectproxy/routes.go#L135) ya'll create, because that would've just let us wrap only a small portion of the logic rather than having to completely reimplement a ResponseWriter to filter requests.

PR #70, opened 2021-06-17
Fully support POST: enforceLabel/matcher/labelAPI (#55)

hanjm:
Closes #55

Grafana always use POST when grafana alert query Prometheus ref: [grafana code](https://github.com/grafana/grafana/blob/main/pkg/tsdb/prometheus/prometheus.go#L89) , [prometheus client_golang code](https://github.com/prometheus/client_golang/blob/master/api/prometheus/v1/api.go#L1101)

Grafana 8.0 default use POST in Prometheus datasource. ref: [grafana#34599](https://github.com/grafana/grafana/pull/34599)

Current, `prom-label-proxy` cannot proxy POST method correctly.  

```
 go run main.go  -label namespace -upstream http://127.0.0.1:9090 -insecure-listen-address :29192 -enable-label-apis -unsafe-passthrough-paths /api/v1/metadata,/api/v1/query_exemplars
```

main/master branch:
image
image

this PR fixed it, make `prom-label-proxy` proxy POST method correctly: 
image

hi @bwplotka  @s-urbaniak Could you help review? thank you :)

PR #85, opened 2021-09-17
always parse request post form

yeya24:
Signed-off-by: Ben Ye <ben.ye@bytedance.com>

Always parse form when we need to access the request body.


yeya24:
cc @hanjm as you wrote this part of code https://github.com/prometheus-community/prom-label-proxy/pull/70/files.

squat (approved):

hanjm:
LGTM. Explicit is clear，though net/http will auto parseForm when call Formvalue. https://github.com/golang/go/blob/master/src/net/http/request.go#L1346

yeya24:
Yeah. As mentioned by @hanjm, looks like we can remove all the `ParseForm` calls because we already parse it at the beginning using `FormValue`.

Issue #51, opened 2020-10-26
Return error response in case of errors


bwplotka:
Wonder if we are doing right think on error case :thinking: 

```
func (r *routes) query(w http.ResponseWriter, req *http.Request) {
	expr, err := parser.ParseExpr(req.FormValue("query"))
	if err != nil {
		return
	}

	e := NewEnforcer([]*labels.Matcher{
		{
			Name:  r.label,
			Type:  labels.MatchEqual,
			Value: mustLabelValue(req.Context()),
		},
	}...)
	if err := e.EnforceNode(expr); err != nil {
		return
	}

	q := req.URL.Query()
	q.Set("query", expr.String())
	req.URL.RawQuery = q.Encode()

	r.handler.ServeHTTP(w, req)
}
```

Shouldn't we do some http.Error at least? :thinking:

simonpasquier:
I thought the same while reviewing your PR. Maybe @brancz @s-urbaniak have more insights?

s-urbaniak:
I believe I stumbled across this quite some time ago and I agree it is non obvious what happens, but effectively currently a parser failure will:

- return http status code 200 (debatable)
- will return an empty body (debatable)
- will not execute an upstream query (good)

We could return StatusBadRequest (400) instead, wdyt?

But somehow I have a dejavu as I am sure I asked this question OOB to @brancz :smile:

brancz:
Error on parsing should be a 400, error on enforcing should be a 500 I think.

jan--f:
Does #67 (at least partially) address this?

PR #97, opened 2021-12-13
fix: returning correct http status code in case of failures

prashbnair:
Fixes  https://github.com/prometheus-community/prom-label-proxy/issues/51

Signed-off-by: Prashant Balachandran <pnair@redhat.com>

### DIFF
--- a/injectproxy/routes.go
+++ b/injectproxy/routes.go
@@ -75,7 +75,7 @@ func WithPassthroughPaths(paths []string) Option {
 	})
 }
 
-// ErrorOnReplace causes the proxy to return 403 if a label matcher we want to
+// WithErrorOnReplace causes the proxy to return 400 if a label matcher we want to
 // inject is present in the query already and matches something different
 func WithErrorOnReplace() Option {
 	return optionFunc(func(o *options) {
@@ -294,8 +294,13 @@ func (r *routes) query(w http.ResponseWriter, req *http.Request) {
 	// enforce in both places.
 	q, found1, err := enforceQueryValues(e, req.URL.Query())
 	if err != nil {
-		if _, ok := err.(IllegalLabelMatcherError); ok {
+		switch err.(type) {
+		case IllegalLabelMatcherError:
 			http.Error(w, err.Error(), http.StatusBadRequest)
+		case queryParseError:
+			http.Error(w, err.Error(), http.StatusBadRequest)
+		case enforceLabelError:
+			http.Error(w, err.Error(), http.StatusInternalServerError)
 		}
 		return
 	}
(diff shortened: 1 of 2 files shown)


injectproxy/routes.go
```diff
@@ -405,3 +420,27 @@ func matchersToString(ms ...*labels.Matcher) string {
...
+type QueryParseError struct {
+	msg string
+}
+
+func (e QueryParseError) Error() string {
+	return e.msg
+}
+
+func newQueryParseError(err error) QueryParseError {
+	return QueryParseError{msg: fmt.Sprintf("Error parsing query string %s", err.Error())}
```
squat, line 433: Perhaps we should quote here to clarify where exactly the query starts and ends
```suggestion
	return QueryParseError{msg: fmt.Sprintf("Error parsing query string %q", err.Error())}
```

squat: Error strings typically start with lower case letters so they can be wrapped:
```suggestion
	return QueryParseError{msg: fmt.Sprintf("error parsing query string %s", err.Error())}
```

injectproxy/routes.go
```diff
@@ -405,3 +420,27 @@ func matchersToString(ms ...*labels.Matcher) string {
...
+type EnforceLabelError struct {
+	msg string
+}
+
+func (e EnforceLabelError) Error() string {
+	return e.msg
+}
+
+func newEnforceLabelError(err error) EnforceLabelError {
+	return EnforceLabelError{msg: fmt.Sprintf("Error enforcing label %s", err.Error())}
```
squat, line 445: Here too:
```suggestion
	return EnforceLabelError{msg: fmt.Sprintf("error enforcing label %q", err.Error())}
```

injectproxy/routes.go
```diff
@@ -405,3 +420,27 @@ func matchersToString(ms ...*labels.Matcher) string {
...
+
+func (e QueryParseError) Error() string {
+	return e.msg
+}
+
+func newQueryParseError(err error) QueryParseError {
+	return QueryParseError{msg: fmt.Sprintf("Error parsing query string %s", err.Error())}
+}
+
+type EnforceLabelError struct {
```
squat, line 436: Does this need to be exported or can we keep it lower case? If it really must be exported, then please add comment to document it
```suggestion
type enforceLabelError struct {
```

injectproxy/routes.go
```diff
@@ -405,3 +420,27 @@ func matchersToString(ms ...*labels.Matcher) string {
 	}
 	return fmt.Sprintf("{%v}", strings.Join(el, ","))
 }
+
+type QueryParseError struct {
```
squat, line 424: Here too re: exporting
```suggestion
type queryParseError struct {
```

squat (commented):
Thanks for doing this @prashbnair, it's long overdue. I just have a few comments

injectproxy/routes.go
prashbnair: Done

injectproxy/routes.go
prashbnair: Done

injectproxy/routes.go
prashbnair: Done

injectproxy/routes.go
prashbnair: Done

simonpasquier (approved):
lgtm

PR #98, opened 2021-12-15
*: bump Prometheus dependency to v2.32.0

simonpasquier:
github.com/prometheus/prometheus/pkg/* packages have been moved to github.com/prometheus/prometheus/model/*.
Since prometheus operator depends also on prom-label-proxy, we need this change before we can bump the prometheus version in the operator repository.


squat (approved):
LGTM

simonpasquier:
:heart:

Issue #27, opened 2020-04-16
change injected PromQL label from labels.MatchEqual to labels. MatchRegexp to support multiple namespaces


dongyingbo:
## Problem
This proxy is used together with kube-rbac-proxy for OCP monitoring multiple tenancy.   

kube-rbac-proxy supports to validate user against multiple namespaces via query like `namespace=kube-system&namespace=openshift-monitoring`.
But this proxy support only one namespace. With the query above, its source code,
```
func (r *routes) ServeHTTP(w http.ResponseWriter, req *http.Request) {
	lvalue := req.URL.Query().Get(r.label)
        ....
}       
```
will extract only namespace `kube-system`.
And source code,
```
	err = SetRecursive(expr, []*labels.Matcher{
		{
			Name:  r.label,
			Type:  labels.MatchEqual,
			Value: mustLabelValue(req.Context()),
		},
	})

```
will add labels.MatchEqual.

## Suggestion
1. In the first piece of source code, extract all query values and construct a RegExpr string like `kube-system|openshift-monitoring`.
2. In the second piece of source code, use labels.MatchRegexp instead of labels.MatchEqual

dongyingbo:
/cc simonpasquier
May you take a look when you are available?

lilic:
Thanks for your issue! Sorry for the late reply. I think having multi namespace support is a good idea, I know that kube-rbac-proxy has this already so maybe Frederic has some ideas around how this should look like in prom-label-proxy as well.

cc @brancz @simonpasquier

brancz:
If multiple values is the use case, I would prefer if we made the query parameter repeatable instead of using a regex. The regex is going to be difficult security wise. Repeated namespaces would have the same effect though, so that should work equally well. I'm happy to have that addition if it doesn't already work.

simonpasquier:
Regex matchers _should_ be safe if we escape values with [`regex.QuoteMeta`](https://golang.org/pkg/regexp/#QuoteMeta) though?

dongyingbo:
Thanks for your reply @lilic @brancz  and @simonpasquier !

I realized that this proxy does not meet our requirements for building Grafana on top of it. The main problem is that I have to update the PromQL string from Grafana before it is passing onto kube-rbac-proy.

So I created one new project and initial source code is uploaded here https://github.com/IBM/ibm-grafana-ocpthanos-proxy. 
The proxy will be used as sidecar of Grafana and it will update PromQL string directly.

[...]

PR #100, opened 2021-12-30
Added support for regex label

NikWaltz:
This PR about support regex label #27. 
Main idea is to recognise matcher from query and keep matcher instead label in context.
It allow to use multiple namespaces via query like `namespace=~kube-system|openshift-monitoring` and still works with one namespace without regex.

### DIFF
--- a/injectproxy/routes.go
+++ b/injectproxy/routes.go
@@ -198,17 +198,46 @@ func NewRoutes(upstream *url.URL, label string, opts ...Option) (*routes, error)
 
 func (r *routes) enforceLabel(h http.HandlerFunc) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, req *http.Request) {
+		queryLabel := r.label
 		lvalue := req.FormValue(r.label)
+		matcherType := labels.MatchEqual
+		if strings.HasPrefix(lvalue, "~") {
+			matcherType = labels.MatchRegexp
+			lvalue = strings.TrimPrefix(lvalue, "~")
+		}
+		if lvalue == "" {
+			queryLabel = r.label + "!"
+			lvalue = req.FormValue(queryLabel)
+			matcherType = labels.MatchNotEqual
+			if lvalue == "" {
+				queryLabel = r.label + "!~"
+				for k, v := range req.Form {
+					if strings.HasPrefix(k, queryLabel) {
+						if v[0] == "" {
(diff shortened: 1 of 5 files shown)


injectproxy/routes.go
```diff
@@ -198,17 +198,46 @@ func NewRoutes(upstream *url.URL, label string, opts ...Option) (*routes, error)
...
+		}
+		if lvalue == "" {
+			queryLabel = r.label + "!"
+			lvalue = req.FormValue(queryLabel)
+			matcherType = labels.MatchNotEqual
+			if lvalue == "" {
+				queryLabel = r.label + "!~"
+				for k, v := range req.Form {
+					if strings.HasPrefix(k, queryLabel) {
+						if v[0] == "" {
```
m-yosefpor, line 216: I think it should be 

```suggestion
						if v[0] != "" {
```

injectproxy/routes.go
```diff
@@ -198,17 +198,46 @@ func NewRoutes(upstream *url.URL, label string, opts ...Option) (*routes, error)
...
+							lvalue = strings.TrimPrefix(k, queryLabel)
```
m-yosefpor, line 217:
```suggestion
							lvalue = strings.TrimPrefix(v[0], queryLabel)
```

m-yosefpor:
query parameter for different types:
👍 

a bit tricky :)
```
q.Add("namespace", "monitoring") // MatchEqual
q.Add("namespace", "~monitoring|default") // MatchRegexp
q.Add("namespace=", "!monitoring") // MatchNotEqual
q.Add("namespace!~monitoring|default", "") // MatchNotRegexp: empty value, all in keys 
```

It makes more sense when converted to query URL :)

m-yosefpor:
> namespace

I think it makes more sens to have 
```
q.Add("namespace, "!~monitoring|default")
```
as match not regexp. The syntax then would be `namespace=!~monitoring|default` but it makes more sense when having query added via function, and also can be used together with features such as `https://github.com/prometheus-community/prom-label-proxy/pull/116` -label-value

sathieu:
See somewhat related https://github.com/prometheus-community/prom-label-proxy/pull/171.

simonpasquier:
From a quick look, #171 should achieve the same goal. Closing this one as it has merge conflicts anyway.

PR #106, opened 2022-06-01
feat: implement healthcheck for kubernetes probes

jkroepke:
Alternate to TCP based probes for kubernetes

### DIFF
--- a/injectproxy/routes.go
+++ b/injectproxy/routes.go
@@ -15,6 +15,7 @@ package injectproxy
 
 import (
 	"context"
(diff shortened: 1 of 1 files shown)


squat (approved):
Sounds nice :)

I'm working on a PR to actually instrument the prom-label-proxy and expose metrics on an internal HTTP server. Once that's in place, we should probably move this handler there, however this is fine for now :+1:

jkroepke:
Thanks @squat . I would really appreciate it, if this could be released soon. 👍

PR #115, opened 2022-06-17
feat: enable multiple label values

Kirchen99:
Signed-off-by: Kirchen99 <latias_latios@126.com>

Solves the issue: https://github.com/prometheus-community/prom-label-proxy/issues/27

Example:

Run prom-label-proxy with following args:
```
                "--insecure-listen-address=127.0.0.1:8080",
                "--upstream=http://demo.do.prometheus.io:9090",
                "--label=job"
```

Try to get something with `curl`:
```
curl http://127.0.0.1:8080/api/v1/query\?query\="up"\&job\="grafana"\&job\="random"\&job\="caddy"
```

See the result:
```
{
   "status":"success",
   "data":{
      "resultType":"vector",
      "result":[
         {
            "metric":{
               "__name__":"up",
               "instance":"demo.do.prometheus.io:3000",
               "job":"grafana"
            },
            "value":[
               1655460956.224,
               "1"
            ]
         },
         {
            "metric":{
               "__name__":"up",
               "instance":"demo.do.prometheus.io:8996",
               "job":"random"
            },
            "value":[
               1655460956.224,
               "1"
            ]
         },
         {
            "metric":{
               "__name__":"up",
               "instance":"demo.do.prometheus.io:8997",
               "job":"random"
            },
            "value":[
               1655460956.224,
               "1"
            ]
         },
         {
            "metric":{
               "__name__":"up",
               "instance":"demo.do.prometheus.io:8998",
               "job":"random"
            },
            "value":[
               1655460956.224,
               "1"
            ]
         },
         {
            "metric":{
               "__name__":"up",
               "instance":"demo.do.prometheus.io:8999",
               "job":"random"
            },
            "value":[
               1655460956.224,
               "1"
            ]
         },
         {
            "metric":{
               "__name__":"up",
               "instance":"localhost:2019",
               "job":"caddy"
            },
            "value":[
               1655460956.224,
               "1"
            ]
         }
      ]
   }
}
```


injectproxy/routes.go
```diff
@@ -205,11 +205,14 @@ func NewRoutes(upstream *url.URL, label string, opts ...Option) (*routes, error)
 
 func (r *routes) enforceLabel(h http.HandlerFunc) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, req *http.Request) {
-		lvalue := req.FormValue(r.label)
-		if lvalue == "" {
+		req.ParseForm()
+		lvalues, present := req.Form[r.label]
+		if !present || len(lvalues) == 0 || lvalues[0] == "" {
```
squat, line 210: should we check if _every_ value is non-empty?

Kirchen99: Yes, that make sense. I'll update the code.

Kirchen99: Or maybe we just make sure, *not all* of the values are empty.

Kirchen99: I have edited the code. Now it checks *every* value, and reports error if one of the value is empty.

injectproxy/routes.go
```diff
@@ -291,7 +307,7 @@ func (r *routes) query(w http.ResponseWriter, req *http.Request) {
 	e := NewEnforcer(r.errorOnReplace,
 		[]*labels.Matcher{{
 			Name:  r.label,
-			Type:  labels.MatchEqual,
+			Type:  labels.MatchRegexp,
```
squat, line 310: Do we need to do any special treatment of the labels so that someone doesn't intentionally insert a regular expression to DOS the proxy, e.g. with lots of backtracking?

Kirchen99: @squat Do you have any example regex expression which could lead to a DOS for testing?

.gitignore
```diff
@@ -1,4 +1,5 @@
 prom-label-proxy
 
 .idea
-.envrc
\ No newline at end of file
+.envrc
+.vscode
```
simonpasquier, line 5: could you remove this change?
self-note: we should remove .envrc and .idea too.

injectproxy/routes.go
```diff
@@ -206,11 +206,27 @@ func NewRoutes(upstream *url.URL, label string, opts ...Option) (*routes, error)
...
-		lvalue := req.FormValue(r.label)
-		if lvalue == "" {
+		err := req.ParseForm()
+		if err != nil {
+			prometheusAPIError(w, fmt.Sprintf("Bad request. The query parameter can not be parsed. %s", err.Error()), http.StatusBadRequest)
+			return
+		}
+
+		lvalues, present := req.Form[r.label]
+		if !present || len(lvalues) == 0 {
```
simonpasquier, line 216:
```suggestion
		lvalues, _ := req.Form[r.label]
		if len(lvalues) == 0 {
```

injectproxy/routes.go
simonpasquier: Shouldn't we use [QuoteMeta](https://pkg.go.dev/regexp#QuoteMeta) to avoid any risk like this?

simonpasquier (commented):
you'd need a rebase :)

injectproxy/routes.go
squat: Yes exactly, that's perfect @simonpasquier this avoids the issue i was worried about

Kirchen99: > Shouldn't we use [QuoteMeta](https://pkg.go.dev/regexp#QuoteMeta) to avoid any risk like this?

Thank you for your feedback. I added a QuoteMeta in this line according to your suggestion. I hope I understand you correctly.

https://github.com/prometheus-community/prom-label-proxy/pull/115/files#diff-d120278b267943f472f753843c2abce05ac45a07a1e53b512dfff1438125b980R284

.gitignore
Kirchen99: Thank you for the feedback. I removed all of them(.envrc, .idea and .vscode)

injectproxy/routes.go
```diff
@@ -260,20 +261,36 @@ func (r *routes) enforceLabel(h http.HandlerFunc) http.Handler {
...
+	for i := range formValues {
+		if formValues[i] == "" {
+			return "", fmt.Errorf("the %q query parameter must be provided", r.label)
+		}
+
+		formValues[i] = regexp.QuoteMeta(formValues[i])
+	}
+
 	if r.labelValue != "" {
 		return r.labelValue, nil
```
simonpasquier, line 288: I'm not sure that we should bypass regexp quoting when `-label-value` is defined.

In the current PR version, you could match on multiple static values doing:

```console
./prom-label-proxy -label project -label-value 'prometheus|alertmanager'
```

But I feel that the following would be safer and more consistent with the "label value via query parameter" mode

```console
./prom-label-proxy -label project -label-value prometheus -label-value alertmanager
```

injectproxy/routes.go
```diff
@@ -260,20 +261,36 @@ func (r *routes) enforceLabel(h http.HandlerFunc) http.Handler {
 // sent through a URL parameter.
 // It returns an error when either the value is found in both places, or is not found at all.
 func (r *routes) getLabelValue(req *http.Request) (string, error) {
-	formValue := req.FormValue(r.label)
-	if r.labelValue != "" && formValue != "" {
+	err := req.ParseForm()
+	if err != nil {
+		return "", fmt.Errorf("the query parameter can not be parsed. %s", err.Error())
```
simonpasquier, line 266:
```suggestion
		return "", fmt.Errorf("the query parameter can not be parsed: %w", err)
```

injectproxy/routes.go
Kirchen99: > I'm not sure that we should bypass regexp quoting when `-label-value` is defined.
> 
> In the current PR version, you could match on multiple static values doing:
> 
> ```
> ./prom-label-proxy -label project -label-value 'prometheus|alertmanager'
> ```
> 
> But I feel that the following would be safer and more consistent with the "label value via query parameter" mode
> 
> ```
> ./prom-label-proxy -label project -label-value prometheus -label-value alertmanager
> ```

I agree with you completely and have changed the code to the consistent way.

injectproxy/routes.go
```diff
@@ -256,24 +257,49 @@ func (r *routes) enforceLabel(h http.HandlerFunc) http.Handler {
...
+		}
+	}
+
+	return r.joinMultipleLabelValues(formValues), nil
+}
+
+func (r *routes) joinMultipleLabelValues(labelValues []string) string {
+	for i := range labelValues {
+		labelValues[i] = regexp.QuoteMeta(labelValues[i])
 	}
```
simonpasquier, line 298: this would modify the given slice which we don't want, especially in the case of the static label values.

```suggestion
        lvs := make([]string, len(labelValues)
	for i := range labelValues {
		lvs[i] = regexp.QuoteMeta(labelValues[i])
	}
```

injectproxy/routes.go
```diff
@@ -256,24 +257,49 @@ func (r *routes) enforceLabel(h http.HandlerFunc) http.Handler {
...
 
-	return formValue, nil
+	lvalue := strings.Join(labelValues, "|")
+
+	return lvalue
```
simonpasquier, line 302:
```suggestion
	return strings.Join(lvs, "|")
```

main.go
```diff
@@ -44,7 +63,7 @@ func main() {
 	flagset.StringVar(&label, "label", "", "The label to enforce in all proxied PromQL queries. "+
 		"This label will be also required as the URL parameter to get the value to be injected. For example: -label=tenant will"+
 		" make it required for this proxy to have URL in form of: <URL>?tenant=abc&other_params...")
-	flagset.StringVar(&labelValue, "label-value", "", "A fixed label value to enforce in all proxied PromQL queries. "+
+	flagset.Var(&labelValue, "label-value", "A fixed label value to enforce in all proxied PromQL queries. "+
```
simonpasquier, line 66:
```suggestion
	flagset.Var(&labelValue, "label-value", "A fixed label value to enforce in all proxied PromQL queries. It can be repeated in which case the proxy will enforce the union of values. "+
```

injectproxy/routes_test.go
```diff
@@ -671,18 +671,19 @@ func TestSeriesWithPost(t *testing.T) {
...
-		promQueryBody  string
-		method         string
-
-		expCode          int
-		expPromQuery     string
-		expPromQueryBody string
-		expResponse      []byte
-		errorOnReplace   bool
+		name                 string
+		labelv               string
```
simonpasquier, line 675: what about using a string slice instead of `secondLabelv` and `forceAddSecondLabelv`?
```suggestion
		labelv               []string
```

Kirchen99: Yes, really elegant solution

injectproxy/routes.go
```diff
@@ -38,15 +39,15 @@ type routes struct {
...
+	labelValue []string
 
 	mux            *http.ServeMux
 	modifiers      map[string]func(*http.Response) error
 	errorOnReplace bool
 }
 
 type options struct {
-	labelValue       string
+	labelValue       []string
```
simonpasquier, line 50:
```suggestion
	labelValues       []string
```

injectproxy/routes.go
```diff
@@ -88,7 +89,7 @@ func WithErrorOnReplace() Option {
 
 // WithLabelValue enforces a specific value for the multi-tenancy label.
 // If not specified, the value has to be provided as a URL parameter.
-func WithLabelValue(value string) Option {
+func WithLabelValue(value []string) Option {
```
simonpasquier, line 92: The comment would also need an update.
```suggestion
func WithLabelValues(value []string) Option {
```

injectproxy/routes.go
```diff
@@ -256,24 +257,48 @@ func (r *routes) enforceLabel(h http.HandlerFunc) http.Handler {
...
 
-// getLabelValue returns the statically set label value, or the label value
-// sent through a URL parameter.
-// It returns an error when either the value is found in both places, or is not found at all.
+// getLabelValue returns a regular expression concatenated using "|" by the statically set label value,
+// or by the label value sent through a URL parameter.
+// It returns an error when the URL parameter cannot be parsed,
+// or the value is found in both places, or is not found at all,
+// or one of the provided parameters is empty
 func (r *routes) getLabelValue(req *http.Request) (string, error) {
```
simonpasquier, line 265: At this stage, we should return a slice of strings because not all proxied endpoints work with regexp matchers (e.g. /api/v1/rules)

```suggestion
func (r *routes) getLabelValues(req *http.Request) ([]string, error) {
```

simonpasquier (commented):
I realize that there's more work needed: while the PR works the /api/v1/query and /api/v1/query_range endpoints, it would break other endpoints such as /federate, /api/v1/rules, /api/v2/silences, ...
I think that `enforceLabel()` needs to store the slice of label values in the context. Then each downstream handler does whatever it needs with it.

injectproxy/routes.go
Kirchen99: I have updated the code. The slice of label values are stored in the context. And in `query()`, `matcher()` and `passthrough()` the context will be processed in different way.

injectproxy/routes.go
```diff
@@ -306,18 +322,40 @@ type ctxKey int
...
 func mustLabelValue(ctx context.Context) string {
-	label, ok := ctx.Value(keyLabel).(string)
+	labels, ok := ctx.Value(keyLabel).([]string)
+	if !ok {
+		panic(fmt.Sprintf("can't find the %q value in the context", keyLabel))
+	}
+	if labels[0] == "" {
+		panic(fmt.Sprintf("empty %q value in the context", keyLabel))
+	}
+	return labels[0]
```
simonpasquier, line 332: I think that it's misleading and introduce inconsistency (e.g. query endpoints support enforcing multiple label values while /api/v1/silences endpoint doesn't as it only looks at the first label value). I would turn `func mustLabelValue(ctx context.Context) string` into `func mustLabelValues(ctx context.Context) []string` and update everywhere it needs to be.

In practice we need to modify

https://github.com/prometheus-community/prom-label-proxy/blob/22ccf0e32939e95ae12187c472e03afed68b7eff/injectproxy/silences.go#L46

https://github.com/prometheus-community/prom-label-proxy/blob/22ccf0e32939e95ae12187c472e03afed68b7eff/injectproxy/silences.go#L75

https://github.com/prometheus-community/prom-label-proxy/blob/22ccf0e32939e95ae12187c472e03afed68b7eff/injectproxy/silences.go#L132

https://github.com/prometheus-community/prom-label-proxy/blob/22ccf0e32939e95ae12187c472e03afed68b7eff/injectproxy/rules.go#L200

https://github.com/prometheus-community/prom-label-proxy/blob/22ccf0e32939e95ae12187c472e03afed68b7eff/injectproxy/rules.go#L226

Kirchen99: @simonpasquier 
I'm now trying to keep every consistent(every endpoints should support multiple label values) and there is something which I can't understand. Could you please explain to me:

Why is the first `namespace="default"` not thrown away?

https://github.com/prometheus-community/prom-label-proxy/blob/1975124010c20f5e227ea7351e0bdea798badbd5/injectproxy/routes_test.go#L288-L295

Kirchen99: Every endpoint can accept multiple label values now. One big change I need to mention is `/api/v2/silences` for `GET`. As I want to keep the consistency of each endpoint that it should work as `union` of all label values, I changed the procedure how the proxy filters the result. It filters the result after it received response from alertmanager, not before it sends request to alertmanager.

Please carefully review this function:
https://github.com/prometheus-community/prom-label-proxy/blob/0471c290c7f24096653d5a37e1a6a021de6ccf73/injectproxy/silences.go#L258-L311

What I want to do here is:

If we have label `namespace` and  values: `["namespace1", "namespace2", "namespace3"]`
All silences with one of the following matchers should be listed with a `GET` request:
```
# non regex:
namespace="namespace1"
namespace="namespace2"
namespace="namespace3"

# regex:
namespace=~"namespace1"
namespace=~"namespace2"
namespace=~"namespace3"
namespace=~"namespace1|namespace2"
namespace=~"namespace1|namespace3"
namespace=~"namespace2|namespace3"
namespace=~"namespace1|namespace2|namespace3"
```

And also be able to be updated/deleted with a `POST` or `DELETE` request. Because we want to have `union`.

Please tell me if there is better way to do this or I think wrong.

injectproxy/routes.go
```diff
@@ -292,24 +294,39 @@ func (r *routes) enforceLabel(h http.HandlerFunc) http.Handler {
...
-		return "", fmt.Errorf("a static value for the %s label has already been specified", r.label)
+// getLabelValues returns a slice of strings of the statically set label value,
+// or of the label value sent through a URL parameter.
+// It returns an error when the URL parameter cannot be parsed,
+// or the value is found in both places, or is not found at all,
+// or one of the provided parameters is empty.
+func (r *routes) getLabelValues(req *http.Request) ([]string, error) {
+	err := req.ParseForm()
+	if err != nil {
+		return []string{}, fmt.Errorf("the query parameter can not be parsed: %w", err)
```
simonpasquier, line 305: (nit) and the same for all error returns

```suggestion
		return nil, fmt.Errorf("the query parameter can not be parsed: %w", err)
```

injectproxy/routes.go
```diff
@@ -249,22 +250,23 @@ func NewRoutes(upstream *url.URL, label string, opts ...Option) (*routes, error)
...
 }
 
 func (r *routes) enforceLabel(h http.HandlerFunc) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, req *http.Request) {
-		lvalue, err := r.getLabelValue(req)
+		lvalues, err := r.getLabelValues(req)
 		if err != nil {
 			prometheusAPIError(w, humanFriendlyErrorMessage(err), http.StatusBadRequest)
 			return
 		}
```
simonpasquier, line 267:
```suggestion
		}
		// Ensure that label values are always sorted in the same order.
		sort.Strings(lvalues)
```

injectproxy/rules.go
```diff
@@ -208,7 +208,7 @@ func (r *routes) filterRules(lvalue string, resp *apiResponse) (interface{}, err
 		var rules []rule
 		for _, rule := range rg.Rules {
 			for _, lbl := range rule.Labels() {
-				if lbl.Name == r.label && lbl.Value == lvalue {
+				if lbl.Name == r.label && contains(lvalues, lbl.Value) {
```
simonpasquier, line 211: It could use https://pkg.go.dev/sort#SearchStrings
```suggestion
				if lbl.Name == r.label && sort.SearchStrings(lvalues, lbl.Value) {
```

simonpasquier (commented):
Thanks a lot for the effort! It's definitely more involved than initially envisioned :sweat_smile: 
I'm still not decided on the best way to deal with the silences endpoint, i'll need to think about it :thinking:

injectproxy/rules.go
Kirchen99: > SearchStrings searches for x in a sorted slice of strings and returns the index as specified by Search. The return value is the index to insert x if x is not present (it could be len(a)). 

If we want to find whether a string is in the slice, it must test `lvalues[i] == lbl.Value` separately.

It is still needed to define a function to deal with it. The function will look like something like this:

```golang
func contains(someValues []string, value string) bool {
    i := sort.SearchStrings(someValues, value)
    
    return i < len(someValues) && someValues[i]==value
}
```

Kirchen99: I have change the code like this:
https://github.com/prometheus-community/prom-label-proxy/blob/95871f0751cae73bde76a8315ac1fc2b7a81a9d1/injectproxy/rules.go#L246-L250

Please tell me if I think wrong.

injectproxy/silences.go
```diff
@@ -90,15 +160,15 @@ func (r *routes) postSilence(w http.ResponseWriter, req *http.Request) {
...
 	}
 
-	var falsy bool
+	truthy := true
 	modified := models.Matchers{
-		&models.Matcher{Name: &(r.label), Value: &lvalue, IsRegex: &falsy},
+		&models.Matcher{Name: &(r.label), Value: &matcherValue, IsRegex: &truthy},
 	}
 	for _, m := range sil.Matchers {
 		if m.Name != nil && *m.Name == r.label {
```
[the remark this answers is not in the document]
Kirchen99: I have updated the code.

If the enforced label is `namespace=~"ns1"`, and someone creates a silence with `namespace="ns1"`, then both matchers will be added: 
```json
[
    "namespace=~\"ns1\"",
    "namespace=\"ns1\""
]
```

For List, Update and Delete, only silences exactly contain a  `namespace=~"ns1"` will be processed.

Did I understand you correctly?

Kirchen99:
I resolved some conversations to make this thread(pull request) easier to read and follow. Please tell me if I'd better not to resolve them. @simonpasquier

gasmick:
Wow, it would be great to have this merged soon!

bt909:
Nice useful feature, is there any plan to merge this soon?

kekscode:
@simonpasquier can we expect that this gets merged anytime soon or are there any open issues you see?

lud97x:
+1

dbluxo:
@simonpasquier doesn't seem to have had time to review this PR further. Would maybe someone else from the maintainers have time to take a look here? @squat @brancz

simonpasquier:
hey! Sorry for the lag on this PR. I'm putting it at the top of my review list as I'm aware that there's high demand.

simonpasquier:
@Kirchen99  now that #118 is merged, you'd have to rebase and resolve conflicts. Sorry about that but once it's done, I'll review the PR :)

Kirchen99:
> @Kirchen99 now that #118 is merged, you'd have to rebase and resolve conflicts. Sorry about that but once it's done, I'll review the PR :)

@simonpasquier I have rebased :) Please review.

README.md
```diff
@@ -105,6 +112,13 @@ prom-label-proxy \
 {"status":"success","data":{"resultType":"vector","result":[]}}%
 ```
 
+You can provide multiple values for the label using http header:
```
simonpasquier, line 115:
```suggestion
You can provide multiple values for the label using several HTTP headers:
```

injectproxy/routes.go
```diff
@@ -204,17 +206,30 @@ func (hff HTTPFormEnforcer) ExtractLabel(next http.HandlerFunc) http.Handler {
...
+	formValues := r.Form[hff.ParameterName]
+
+	if len(formValues) == 0 {
+		return nil, fmt.Errorf("the %q query parameter must be provided", hff.ParameterName)
+	}
+
+	// Check if every value is non-empty
+	for i := range formValues {
+		if formValues[i] == "" {
+			return nil, fmt.Errorf("the %q query parameter must be provided", hff.ParameterName)
```
simonpasquier, line 228: I wonder if we shouldn't be liberal and skip empty values instead.

simonpasquier (commented):
thanks for the rebase. I did only a quick review and I'll need to dig deeper into the Alertmanager part but overall, this looks good.

injectproxy/routes.go
Kirchen99: Agreed. I have similar feel that we are too aggressive to report an error here. I have also changed the behavior for static label value and for header. Please have a look as well.

dzabel:
We are looking to use this feature. Any progress?

daskanu:
Can you give us an update to the progress? We also want to use this feature.

Kirchen99:
If I understand correctly, the question is how do we deal with the Alertmanager part. In particular, how we deal with Silence. I'm flexible to accept any idea or requirements to change.

injectproxy/silences.go
```diff
@@ -35,21 +36,76 @@ import (
...
+		}
+
+		if resp.StatusCode != http.StatusOK {
+			// Pass non-200 responses as-is.
+			return nil
+		}
+
+		apir, err := getAlertmanagerAPIResponse(resp.Body)
+		if err != nil {
+			return errors.Wrap(err, "can't decode API response")
```
SuperQ, line 60: Would you mind rebasing this. The `github.com/pkg/errors` errors package is deprecated and archived. You can now do this:

```suggestion
			return fmt.Errorf("can't decode API response: %w", err)
```

SuperQ (changes requested):
Would you mind rebasing this change against the current main?

SuperQ:
Besides the errors thing, this looks good to merge.

simonpasquier:
:wave: I'm taking a look at it right now. Since the silences API is problematic with multiple label values, I'd be in favor of declaring that the proxy doesn't support silences in this mode of operations. The main use case of multi-labels is for the Prometheus API anyway so not supporting the Alertmanager API isn't a blocker IMHO.

I'm going to push some commits on top of your PR to implement this.

simonpasquier:
@Kirchen99 I've added a couple of commits on top of your work. The main changes are
* Any call to the Alertmanager Silences endpoints with multi label values configured returns `422 Unprocessable Entity` (https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/422).
* When multi label values are configured, the proxy won't replace the existing matchers. E.g. when you query `up{namespace="bar"}`, the proxy should translate to `up{namespace="bar",namespace=~"bar|foo"}` and not `up{namespace=~"bar|foo"}`.
* The same applies to the Alertmanager's Alerts and Alert groups API.

Regarding the last 2 points, I don't remember if we already discussed it in this PR but IMHO this is what makes most sense  compared to dropping the existing matcher unilaterally.

My changes are https://github.com/prometheus-community/prom-label-proxy/pull/115/files/8472eba307611d1f2cf9181e70e3319ecbb90976..4882a18523e612324514b1f5ff8c7f8c465e71cc. Happy to hear your feedback.

Kirchen99:
@simonpasquier Thank you for your commits and the detailed explanation of the changes. I like your Idea on the last 2 points and really appreciate your work and the improvements you've made. The changes look good to me, and I'm excited to see them being merged.

SuperQ (approved):

PR #116, opened 2022-06-23
Allow setting static label value on start

fpetkovski:
This commit implements a -label-value flag which allows setting a static
label value when running prom-label-proxy. When this flag is set,
requests which contain a label value in the URL will get rejected
with a bad response status code.

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,7 +75,7 @@ The "tenant" query parameter must be provided.
 {"status":"success","data":{"resultType":"vector","result":[]}}%
 ```
 
-You can also provide a static value for the tenancy label. For example, running `prom-label-proxy` with
+You can also provide a static value for a label. For example, running `prom-label-proxy` with
 ```
 prom-label-proxy \
    -label tenant \
--- a/README.md
+++ b/README.md
@@ -70,10 +70,23 @@ Accessing demo Prometheus APIs on `127.0.0.1:8080` will now expect `tenant` quer
 
 ```bash
 ➜  ~ curl http://127.0.0.1:8080/api/v1/query\?query="up"
-Bad request. The "tenant" query parameter must be provided.
+The "tenant" query parameter must be provided.
 ➜  ~ curl http://127.0.0.1:8080/api/v1/query\?query="up"\&tenant\="something"
 {"status":"success","data":{"resultType":"vector","result":[]}}%
 ```
+
+You can also provide a static value for the tenancy label. For example, running `prom-label-proxy` with
+```
+prom-label-proxy \
+   -label tenant \
+   -value prometheus \
+   -upstream http://demo.do.prometheus.io:9090 \
+   -insecure-listen-address 127.0.0.1:8080
+```
+will enforce `tenant=prometheus` in all requests.
+
+
+In this mode, sending the label value as a query parameter will result in the request getting rejected as a 400 Bad Request.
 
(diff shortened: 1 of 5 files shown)


injectproxy/routes.go
```diff
@@ -206,7 +222,12 @@ func NewRoutes(upstream *url.URL, label string, opts ...Option) (*routes, error)
...
 func (r *routes) enforceLabel(h http.HandlerFunc) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, req *http.Request) {
-		lvalue := req.FormValue(r.label)
+		lvalue, err := r.getLabelValue(req)
+		if err != nil {
+			prometheusAPIError(w, err.Error(), http.StatusBadRequest)
+			return
+		}
+
 		if lvalue == "" {
```
simonpasquier, line 231: I guess we can remove this test since `getLabelValue()` always returns a non-empty string (it should probably be highlighted in the comments and/or function name).

simonpasquier (commented):
Cool! It might be worth highlighting the option in the README.md?

injectproxy/routes.go
fpetkovski: I think it can still return an empty string when both the `r.labelValue` and form value are empty. But we handle the empty case there which would remove the need to check for empty at the call site. Wdyt?

simonpasquier: you're correct, I misread the code :)
what you suggested sounds good to me

injectproxy/routes.go
```diff
@@ -239,6 +256,26 @@ func (r *routes) enforceLabel(h http.HandlerFunc) http.Handler {
...
+// sent through a URL parameter.
+// It returns an error when either the value is found in both places, or is not found at all.
+func (r *routes) getLabelValue(req *http.Request) (string, error) {
+	formValue := req.FormValue(r.label)
+	if r.labelValue != "" && formValue != "" {
+		return "", fmt.Errorf("a static value for the %s label has already been specified", r.label)
+	}
+
+	if r.labelValue == "" && formValue == "" {
+		return "", fmt.Errorf("bad request. The %q query parameter must be provided", r.label)
```
simonpasquier, line 269: I would simplify the error message
```suggestion
		return "", fmt.Errorf("the %q query parameter must be provided", r.label)
```

simonpasquier (commented):
What do you think about updating the README.md?

README.md
```diff
@@ -70,11 +70,24 @@ Accessing demo Prometheus APIs on `127.0.0.1:8080` will now expect `tenant` quer
 
 ```bash
 ➜  ~ curl http://127.0.0.1:8080/api/v1/query\?query="up"
-Bad request. The "tenant" query parameter must be provided.
+The "tenant" query parameter must be provided.
 ➜  ~ curl http://127.0.0.1:8080/api/v1/query\?query="up"\&tenant\="something"
 {"status":"success","data":{"resultType":"vector","result":[]}}%
 ```
 
+You can also provide a static value for the tenancy label. For example, running `prom-label-proxy` with
```
squat, line 78: Tiny nit: we may want to clarify that any label can be set statically, not just a magic tenancy label.
```suggestion
You can also provide a static value for a label. For example, running `prom-label-proxy` with
```

squat (approved):
Looks great to me, @fpetkovski
Just one tiny suggestion from my side

squat (approved):

simonpasquier (approved):
:tada:

PR #118, opened 2022-07-05
feat: Provide label value by http header

### LINKED ISSUE #117: Restrict query by http header as alternative to URL query parameter
Problem:

prom-label-proxy restrict PromQL queries by compare a label value from query with an HTTP GET parameter.

In Grafana, it's possible to define custom additional GET parameters. Sound good.

Today, I figure out that grafana in not enforce the specific get parameter from datasource. Instead Grafana will instruct the local browser to add the custom get parameter to all requests against the datasource. Local modification is possible and users can break out from the jail.

Proposal:

Grafana also support to set custom http headers for datasources. For example grafana use this for multi tenancy queries at Loki.

If the proposal is fine, I can provide a PR to integrate this functionally.

jkroepke:
Fixes #117


kfox1111:
Can we get this merged? Would be very useful.

jkroepke:
@squat could you take a look here?

ThisIsQasim:
Tagging all owners since there have been related PRs (#54 #64 #117 ) with the same feature pending for quite a while.

@brancz @krasi-georgiev @metalmatze @paulfantom @pgier @s-urbaniak @simonpasquier @squat @lilic

jkroepke:
@fpetkovski maybe?

Atomique:
Would also be interested into this. Also into a new release because the actual 0.5.0 doesnt support the -values parameter (only the version in main does this).

injectproxy/routes.go
```diff
@@ -257,23 +258,40 @@ func (r *routes) enforceLabel(h http.HandlerFunc) http.Handler {
 }
 
 // getLabelValue returns the statically set label value, or the label value
-// sent through a URL parameter.
+// sent through a URL parameter or http request header.
 // It returns an error when either the value is found in both places, or is not found at all.
 func (r *routes) getLabelValue(req *http.Request) (string, error) {
+	headerValue, ok := req.Header[headerPrefix+r.label]
```
fpetkovski, line 264: To avoid having deep nesting, maybe we can call one function to get the label value. This function will first use the header, and if it's empty, it will fall back to the query parameter.

Then once we have a value, we can check if an explicit value through `-value` is set and return an error if that's true.

README.md
```diff
@@ -66,13 +66,15 @@ prom-label-proxy \
    -insecure-listen-address 127.0.0.1:8080
 ```
 
-Accessing demo Prometheus APIs on `127.0.0.1:8080` will now expect `tenant` query parameter to be set in the URL:
+Accessing demo Prometheus APIs on `127.0.0.1:8080` will now expect `tenant` query parameter or http header to be set in the URL:
```
simonpasquier, line 69:
```suggestion
Accessing the demo Prometheus APIs on `127.0.0.1:8080` will now expect that the client' request provides the `tenant` label value using either the `tenant` HTTP query parameter or the `x-prom-label-proxy-tenant` HTTP header:
```

simonpasquier (commented):
I'd expect that users pass the name of the header from the command-line (e.g. `--header`). And one of `--query-param` and `--header` needs to be not empty.

For backward-compatibility and principle of least surprise, the default value of `--header` should be "" IMHO. Thus when you upgrade to the latest version, it doesn't become automatically possible to pass the label value via HTTP header (existing setups might not filter HTTP headers so end-users could bypass prom-label-proxy's enforcement).

injectproxy/routes.go
simonpasquier: :+1:

injectproxy/routes.go
```diff
@@ -293,23 +313,54 @@ func (r *routes) enforceLabel(h http.HandlerFunc) http.Handler {
 }
 
 // getLabelValue returns the statically set label value, or the label value
-// sent through a URL parameter.
+// sent through a URL parameter or http request header.
```
simonpasquier, line 316:
```suggestion
// sent through a URL parameter or HTTP request header.
```

main.go
```diff
@@ -48,6 +50,8 @@ func main() {
 	flagset := flag.NewFlagSet(os.Args[0], flag.ExitOnError)
 	flagset.StringVar(&insecureListenAddress, "insecure-listen-address", "", "The address the prom-label-proxy HTTP server should listen on.")
 	flagset.StringVar(&internalListenAddress, "internal-listen-address", "", "The address the internal prom-label-proxy HTTP server should listen on to expose metrics about itself.")
+	flagset.StringVar(&queryParam, "query-param", "", "Name of the GET query param to define the tenant. If empty and header-name is empty, the value of label will be used.")
```
simonpasquier, line 53: The `-label-value` description would need an update to say "the label value will be taken from the URL parameter ... or from the HTTP header ...".
Also it would be good to mark `--label`  as deprecated.

injectproxy/routes.go
```diff
@@ -41,6 +41,8 @@ type routes struct {
 	handler    http.Handler
 	label      string
```
simonpasquier, line 42: For clarity, I would remove the `label` field and rely only on headerName/queryName. Accordingly the NewRoutes() signature would need to be updated to

```
NewRoutes(upstream *url.URL, opts ...Option)
````

And an error returned if none of headerName, queryName or labelValue are set.

File: README.md [hunk 6]
```diff
@@ -70,13 +70,15 @@ prom-label-proxy \
    -insecure-listen-address 127.0.0.1:8080
 ```
 
-Accessing demo Prometheus APIs on `127.0.0.1:8080` will now expect `tenant` query parameter to be set in the URL:
+Accessing the demo Prometheus APIs on `127.0.0.1:8080` will now expect that the client' request provides the `tenant` label value using either the `tenant` HTTP query parameter (configurable by -query-param) or a HTTP header (configurable by -header-name) :
```
simonpasquier, line 73: The command line above needs an update
```
prom-label-proxy \
   -query-param tenant \
   -upstream http://demo.do.prometheus.io:9090 \
   -insecure-listen-address 127.0.0.1:8080
```

simonpasquier (commented):
If we go the route of adding a `--query-param` flag (which is fine by me because more explicit), I think that we should deprecate the `--label` flag.

main.go
jkroepke: > Also it would be good to mark --label as deprecated.

Why? Its still used for configure the restriction label on PromQL side.

injectproxy/routes.go
jkroepke: Not sure, label is still referenced on some places

main.go
simonpasquier: right my bad!

simonpasquier (commented):
@jkroepke sorry for the lag! I have some comments on the CLI help. I've submitted https://github.com/jkroepke/prom-label-proxy/pull/1 to your fork with some refactoring suggestions :)

jkroepke:
Hey @simonpasquier, I saw you suggestions, but I recommend to create an additional PR which superseeds this PR or modify to existing one. "Allow edits by maintainers" is enabled. if you have maintainer permissions on https://github.com/prometheus-community/prom-label-proxy, feel free to commit here. I wont do any refactorings, since I don't have the full golang knowledge.

simonpasquier:
@jkroepke can you reopen my PR on your fork and merge it then? I could have pushed my changes on top of your branch but I didn't want to do it without you being aware :)

jkroepke:
Hey @simonpasquier please take a look that the merge conflict now which appears now after merging the PR

simonpasquier:
@jkroepke sure, let me resolve this!

simonpasquier:
@fpetkovski I'd like your input on this PR. In particular there's a slight difference compared to what you implemented in #116. With this change, users have to select the method to retrieve the label value from incoming requests:
* from the HTTP form/query parameters (`-query-param` or `-label`)
* from the HTTP headers (`-header-name`)
* statically defined (`-label-value`)

It isn't possible to pass `-label-value` and `-label` at the same time hence when a static label value is used, the proxy will not remove anything from the HTTP parameters. Did you have a concrete use case for cleaning up the HTTP parameters when using the static configuration?

simonpasquier:
> It isn't possible to pass `-label-value` and `-label` at the same time hence when a static label value is used, the proxy will not remove anything from the HTTP parameters. Did you have a concrete use case for cleaning up the HTTP parameters when using the static configuration?

After discussing offline with @fpetkovski the statement above isn't correct: in the current version, the proxy will return an HTTP error code when configured with `-label-value` and the incoming HTTP request has a label query parameter. With this PR, the proxy configured with `-label-value` won't remove any HTTP query/form parameter but it sounds ok as it will be ignored anyway.

simonpasquier (approved):

simonpasquier:
thanks a lot @jkroepke for your work and patience!

PR #119, opened 2022-07-07
feat: add /metrics endpoint

jvrplmlmn:
Adds a `/metrics` endpoint to expose default Go metrics.

I initially went for adding this to the existent listener, but if needed be, we can create a dedicated listener for metrics only.


squat (approved):
Thanks @jvrplmlmn :tada: 

This is fine for now, though we will likely want to put this on a dedicated listener, as the prom-label-proxy can be external facing and we shouldn't be sharing all of the metrics publicly.

We will also eventually add HTTP-level metrics about requests handled by the proxy

Issue #135, opened 2023-02-07
[feature request] add support for host header rewrite


FuyuanChu:
Traffic flow:

```
client  --> proxy.local.example.com[prom-label-proxy] ---> central.example.com[upstream]
```

We are using an ingress hostname as the upstream of a `prom-label-proxy`. The ingress relies on the host header to route the traffic.

But currently, the proxy doesn't rewrite the host header. As a result, `prom-label-proxy` will send requests to the upstream with the host header `proxy.local.example.com` and it'll always get a 404 error since the ingress couldn't find a proper route for this hostname.

So we are here to request a feature flag that could rewrite the host header while proxy requests.

PR #154, opened 2023-07-14
Add options to set headers on upstream http requests

graemechristie:
This PR adds the command line option extra-http-header, which facilitates setting arbitrary http headers on the upstream prometheus request. This is required to support queries to prometheus compatible endpoints like Grafana Mimir, which expect an "X-Scope-OrgID:" header in the request to specify the tenant to query the data for.

Additionally, I have added a rewrite-host-header-to parameter, which supports rewriting the host to a different value than the source request.  This would be needed in the case where the upstream prometheus endpoint is an nginx proxy or simliat using host base proxying. This addresses the issue https://github.com/prometheus-community/prom-label-proxy/issues/135.


graemechristie:
Does not seem to be alot of activity in this repo. @SuperQ could you have a look at this PR ?

graemechristie:
@simonpasquier are you able to look this PR ?

injectproxy/routes.go
```diff
@@ -262,6 +282,7 @@ func (sle StaticLabelEnforcer) ExtractLabel(next http.HandlerFunc) http.Handler
 }
 
 func NewRoutes(upstream *url.URL, label string, extractLabeler ExtractLabeler, opts ...Option) (*routes, error) {
+	//extraHttpHeaders []string, rewriteHostHeader string
```
simonpasquier, line 285: to be removed?

main.go
```diff
@@ -81,6 +83,8 @@ func main() {
 		"This option is checked after Prometheus APIs, you cannot override enforced API endpoints to be not enforced with this option. Use carefully as it can easily cause a data leak if the provided path is an important "+
 		"API (like /api/v1/configuration) which isn't enforced by prom-label-proxy. NOTE: \"all\" matching paths like \"/\" or \"\" and regex are not allowed.")
 	flagset.BoolVar(&errorOnReplace, "error-on-replace", false, "When specified, the proxy will return HTTP status code 400 if the query already contains a label matcher that differs from the one the proxy would inject.")
+	flagset.Var(&extraHttpHeaders, "extra-http-header", "Additional HTTP headers to add to the upstream prometheus query in the format 'header: value'. Can be repeated multiple times for additional headers.")
```
simonpasquier, line 86: (nit)
```suggestion
	flagset.Var(&extraHttpHeaders, "extra-http-header", "HTTP header to add to the upstream query in the format 'header: value'. Can be repeated multiple times.")
```

injectproxy/routes.go
```diff
@@ -271,7 +292,19 @@ func NewRoutes(upstream *url.URL, label string, extractLabeler ExtractLabeler, o
...
+	proxy := &httputil.ReverseProxy{
+		Rewrite: func(r *httputil.ProxyRequest) {
+			r.SetURL(upstream)
+			if len(opt.rewriteHostHeader) == 0 {
+				r.Out.Host = r.In.Host
+			} else {
+				r.Out.Host = opt.rewriteHostHeader
+			}
+			for header, val := range opt.extraHttpHeaders {
+				r.Out.Header[header] = []string{val}
```
simonpasquier, line 304: (nit)
```suggestion
				r.Out.Header.Set(header, val)
```

main.go
```diff
@@ -81,6 +83,8 @@ func main() {
...
+	flagset.StringVar(&rewriteHostHeader, "rewrite-host-header-to", "", "Rewrite host header to supplied value when sending the query to the upstream URL.")
```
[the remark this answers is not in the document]
graemechristie: I lodged this PR a long time ago ... and I cannot remember if there was a valid reason for needing the separate argument/behaviour, or I just thought it was preferable to specify the simple rewrite-host-header argument than needing to know the internal mechanics and set the header manually. Regrardless - I primarily need just the extra http headers functionality to support Mimir so I am happy to resubmit this PR with just that option, and we can add the other option in a separate PR if it is required.

injectproxy/routes.go
```diff
@@ -271,7 +286,19 @@ func NewRoutes(upstream *url.URL, label string, extractLabeler ExtractLabeler, o
 		opt.registerer = prometheus.NewRegistry()
 	}
 
-	proxy := httputil.NewSingleHostReverseProxy(upstream)
+	proxy := &httputil.ReverseProxy{
+		Rewrite: func(r *httputil.ProxyRequest) {
+			r.SetURL(upstream)
+			if opt.rewriteHostHeader == "" {
+				r.Out.Host = r.In.Host
```
graemechristie, line 293: OK, I think this is the reason for the separate arguments. I do not think you can set well known mandatory headers like host using headers collection on the request, it has to be set using host property. You *could* technically scan for the key of "host" in the extra headers collection, then if found remove it from the collection and execute this logic .. but that feels hacky. I think the original implementation is the correct behaviour. @simonpasquier - Would you prefer I leave this as is or just remove the ability to set the host header all together ?

main.go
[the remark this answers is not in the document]
graemechristie: @simonpasquier see my comment on line 293 above.

injectproxy/routes.go
graemechristie: I've removed that

injectproxy/routes.go
```diff
@@ -96,6 +98,24 @@ func WithErrorOnReplace() Option {
...
+	return optionFunc(func(o *options) {
+		o.extraHttpHeaders = make(map[string]string)
+		for _, headerArg := range headers {
+			header, val, found := strings.Cut(headerArg, ":")
+			if found {
+				o.extraHttpHeaders[strings.TrimSpace(header)] = strings.TrimSpace(val)
+			}
+		}
+	})
+}
```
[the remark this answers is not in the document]
graemechristie: I don't see how main would have split the value of the each of the string on the ":" ... I've added that logic into main but I thought it would be cleaner here.

main.go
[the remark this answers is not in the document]
graemechristie: So conceptually, "extra-http-headers" are user defined HTTP headers outside of the well known/mandatory headers such as host:.  Thus the -extra-http-headers option only covers these headers in line with how the Go SDK handles these. Host is a specific use case of a mandatory/well known header that needs to be overridden in legitimate scenarios, so that has it's own flag. if we were to allow "Host:" to be overridden with the --extra-http-header flag, we would probably want to override all of the other  mandatory http header edge cases as well, for very little or no benefit.

injectproxy/routes.go
kyrianae: As Host header is very specific, I think it's better to let it alone, and add some treatment (iterate all extra header and check if ...) on each query seems not very optimized

main.go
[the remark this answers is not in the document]
kyrianae: reason to separate Host header come from lines 288-300 from routes.go

kyrianae (approved):
All seems ok to me
I tested this on my own use case and it's working, I did'nt any issue

jellyfish-headphones:
Hello, would love to see this merged. Prom-label-proxy is the exact solution we need for one of our projects. Unfortunately, our mTLS security requirements within our cluster prevent this PLP -> Prometheus networking without the ability to insert a particular header. Would greatly appreciate if someone would help take this to the finish line. Thanks.

dhanushm7:
Hello @graemechristie, this is exactly the solution we needed for our internal use case as well. Would love to see this get merged soon. Thanks!

Issue #82, opened 2021-09-03
Specify label value in prom-label-proxy config / Wildcards in Labels


stieglma:
Hey all,

is it by design, that the value of the label to be checked for cannot be specified directly here, and instead I'll need to put another reverse-proxy between prometheus/thanos and grafana?
I explicitly don't want to use the datasource query parameters in grafana, because grafana users might have the possibility to change that, and therefore they could themselves change to which resources they have access. I also saw the caddy example, and while that would definitely work I think it would be quite nice to not need to start an additional container for that purpose.

Furthermore it would also be quite nice, if the value given for the label could be a wildcard, e.g. on our cluster we have prefixes for certain namespaces that share the same access rights, so ideally i want to set something like `ns-prefix-*` as label value. I also tried with url encoding: `ns-prefix-%2A` but that didn't help either.

If you think these requests are valid, I'm happy to try and add those features as a PR, I did however not work with go up to now, so it might take a while

stieglma:
Here's a first version of the changes: https://github.com/stieglma/prom-label-proxy/commit/fe2275d3bad3198b9ba6a598f09a038ee4632f7c unit-tests still need to be fixed for the MatchRegex feature, and I need to think if there is any potential security risk involved when having that change. Maybe it should just be another setting that can be enabled, something like `--allow-regex`

simonpasquier:
> is it by design, that the value of the label to be checked for cannot be specified directly here, and instead I'll need to put another reverse-proxy between prometheus/thanos and grafana?

Not sure if it's something that we've already evaluated...

> Furthermore it would also be quite nice, if the value given for the label could be a wildcard

The requirement for allowing multiple label values instead of just one has been discussed in #27. Using regular expressions has been dismissed so far because of security implications (too easy for users to get it wrong) but supporting repeated label values would be ok  (see [here](https://github.com/prometheus-community/prom-label-proxy/issues/27#issuecomment-652498598) for more details). I understand that it isn't exactly what you're asking for but would it be enough for your use case? Maybe supporting simple wildcards (e.g. `<prefix>-*`) might be acceptable but I'm curious to hear feedback from other @prom-label-proxy maintainers.

stieglma:
Having repeated label values would work, but needs plenty of more "glue-code" (e.g. when a new namespace with a certain prefix is added, we need to run some kind pipeline which adds another label value, either to the datasource config in grafana, or to our proxy inbetween).

[...]

PR #171, opened 2023-12-07
Accept label value as regexp

sathieu:
Fixes: #82

### DIFF
--- a/README.md
+++ b/README.md
@@ -144,6 +144,30 @@ prom-label-proxy \
 
 `prom-label-proxy` will enforce the `tenant=~"prometheus|alertmanager"` label selector in all requests.
 
+You can match the label value  using a regular expression with the `-regex-match` option. For example:
+
+```
+prom-label-proxy \
+   -label-value '^foo-.+$' \
+   -label namespace \
+   -upstream http://demo.do.prometheus.io:9090 \
+   -insecure-listen-address 127.0.0.1:8080 \
+   -regex-match
+```
+
+> :warning: The above feature is experimental. Be careful when using this option, it may expose sensitive metrics if you use a too permissive expression.
+
+To error out when the query already contains a label matcher that differs from the one the proxy would inject, you can use the `-error-on-replace` option. For example:
+
+```
+prom-label-proxy \
+   -header-name X-Namespace \
+   -label namespace \
+   -upstream http://demo.do.prometheus.io:9090 \
+   -insecure-listen-address 127.0.0.1:8080 \
+   -error-on-replace
+```
+
 Once again for clarity: **this project only enforces a particular label in the respective calls to Prometheus, it in itself does not authenticate or
 authorize the requesting entity in any way, this has to be built around this project.**
 
--- a/injectproxy/routes.go
+++ b/injectproxy/routes.go
@@ -46,6 +46,7 @@ type routes struct {
 	mux            http.Handler
 	modifiers      map[string]func(*http.Response) error
 	errorOnReplace bool
+	regexMatch     bool
 }
 
 type options struct {
@@ -53,6 +54,7 @@ type options struct {
 	passthroughPaths []string
 	errorOnReplace   bool
 	registerer       prometheus.Registerer
+	regexMatch       bool
 }
 
 type Option interface {
@@ -93,6 +95,13 @@ func WithPassthroughPaths(paths []string) Option {
 func WithErrorOnReplace() Option {
 	return optionFunc(func(o *options) {
 		o.errorOnReplace = true
+	})
+}
+
+// WithRegexMatch causes the proxy to handle tenant name as regexp
+func WithRegexMatch() Option {
+	return optionFunc(func(o *options) {
+		o.regexMatch = true
 	})
 }
 
@@ -279,6 +288,7 @@ func NewRoutes(upstream *url.URL, label string, extractLabeler ExtractLabeler, o
 		label:          label,
 		el:             extractLabeler,
 		errorOnReplace: opt.errorOnReplace,
+		regexMatch:     opt.regexMatch,
 	}
 	mux := newStrictMux(newInstrumentedMux(http.NewServeMux(), opt.registerer))
 
@@ -305,18 +315,21 @@ func NewRoutes(upstream *url.URL, label string, extractLabeler ExtractLabeler, o
 		// Reject multi label values with assertSingleLabelValue() because the
 		// semantics of the Silences API don't support multi-label matchers.
 		mux.Handle("/api/v2/silences", r.el.ExtractLabel(
-			enforceMethods(
-				assertSingleLabelValue(r.silences),
-				"GET", "POST",
+			r.errorIfRegexpMatch(
+				enforceMethods(
+					assertSingleLabelValue(r.silences),
+					"GET", "POST",
+				),
 			),
 		)),
 		mux.Handle("/api/v2/silence/", r.el.ExtractLabel(
-			enforceMethods(
-				assertSingleLabelValue(r.deleteSilence),
-				"DELETE",
+			r.errorIfRegexpMatch(
+				enforceMethods(
+					assertSingleLabelValue(r.deleteSilence),
+					"DELETE",
+				),
 			),
 		)),
-
 		mux.Handle("/api/v2/alerts/groups", r.el.ExtractLabel(enforceMethods(r.enforceFilterParameter, "GET"))),
 		mux.Handle("/api/v2/alerts", r.el.ExtractLabel(enforceMethods(r.alerts, "GET"))),
 	)
@@ -386,6 +399,17 @@ func enforceMethods(h http.HandlerFunc, methods ...string) http.HandlerFunc {
 	}
 }
 
+func (r *routes) errorIfRegexpMatch(next http.HandlerFunc) http.HandlerFunc {
+	return func(w http.ResponseWriter, req *http.Request) {
+		if r.regexMatch {
+			prometheusAPIError(w, "support for regex match not implemented", http.StatusNotImplemented)
+			return
+		}
+
+		next(w, req)
+	}
+}
+
 type ctxKey int
 
 const keyLabel ctxKey = iota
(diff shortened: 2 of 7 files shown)


simonpasquier (commented):
Thanks for tackling this! Reading the issue, I have the impression that supporting regexp label values only for `--label-value` was enough (e.g. only for static values). Do you have a use case for supporting the same for `--query-param` and `--header-name`?

Given the security implications (e.g. it's easy to get the regexp wrong and expose too many metrics by accident), I'd be more inclined to limit the scope for now.

README.md
```diff
@@ -144,6 +144,28 @@ prom-label-proxy \
...
+```
+prom-label-proxy \
+   -header-name X-Namespace \
+   -label namespace \
+   -upstream http://demo.do.prometheus.io:9090 \
+   -insecure-listen-address 127.0.0.1:8080 \
+   -error-on-replace
+```
+
+You can match label value  using a regular expression with the `-regex-match` option. For example:
```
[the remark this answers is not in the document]
sathieu: Done

main.go
```diff
@@ -81,6 +82,7 @@ func main() {
 		"This option is checked after Prometheus APIs, you cannot override enforced API endpoints to be not enforced with this option. Use carefully as it can easily cause a data leak if the provided path is an important "+
 		"API (like /api/v1/configuration) which isn't enforced by prom-label-proxy. NOTE: \"all\" matching paths like \"/\" or \"\" and regex are not allowed.")
 	flagset.BoolVar(&errorOnReplace, "error-on-replace", false, "When specified, the proxy will return HTTP status code 400 if the query already contains a label matcher that differs from the one the proxy would inject.")
+	flagset.BoolVar(&regexMatch, "regex-match", false, "When specified, the tenant name is threated as regexp.")
```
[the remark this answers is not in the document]
sathieu: Done.

simonpasquier:
Thanks for the details, it helps a lot! I could agree to the feature being marked as experimental with an very clear warning in the CLI help + README.md file.

The proxy needs also to validate that the regexp is valid. Also what should the proxy's response when it gets several label values? My gut feeling is that it should fail.

README.md
```diff
@@ -144,6 +144,30 @@ prom-label-proxy \
 
 `prom-label-proxy` will enforce the `tenant=~"prometheus|alertmanager"` label selector in all requests.
 
+You can match label value  using a regular expression with the `-regex-match` option. For example:
```
simonpasquier, line 147: (nit)
```suggestion
You can match the label value  using a regular expression with the `-regex-match` option. For example:
```

README.md
```diff
@@ -144,6 +144,30 @@ prom-label-proxy \
...
+
+```
+prom-label-proxy \
+   -header-name X-Namespace \
+   -label namespace \
```
simonpasquier, line 152: (suggestion) I'd use a static label value to make it more "visible".
```suggestion
   -label-value '^foo-.+$' \
   -label namespace \
```

main.go
```diff
@@ -81,6 +82,7 @@ func main() {
 		"This option is checked after Prometheus APIs, you cannot override enforced API endpoints to be not enforced with this option. Use carefully as it can easily cause a data leak if the provided path is an important "+
 		"API (like /api/v1/configuration) which isn't enforced by prom-label-proxy. NOTE: \"all\" matching paths like \"/\" or \"\" and regex are not allowed.")
 	flagset.BoolVar(&errorOnReplace, "error-on-replace", false, "When specified, the proxy will return HTTP status code 400 if the query already contains a label matcher that differs from the one the proxy would inject.")
+	flagset.BoolVar(&regexMatch, "regex-match", false, "When specified, the tenant name is treated as a regular expression.")
```
simonpasquier, line 85:
```suggestion
	flagset.BoolVar(&regexMatch, "regex-match", false, "When specified, the tenant name is treated as a regular expression. In this case, only one tenant name should be provided.")
```

main.go
```diff
@@ -125,6 +127,9 @@ func main() {
 	if errorOnReplace {
 		opts = append(opts, injectproxy.WithErrorOnReplace())
 	}
+	if regexMatch {
+		opts = append(opts, injectproxy.WithRegexMatch())
```
simonpasquier, line 131: in case of static label value (e.g. `-label-value`), we can fail error if several values are provided.

injectproxy/routes.go
```diff
@@ -438,16 +448,28 @@ func (r *routes) query(w http.ResponseWriter, req *http.Request) {
 	var matcher *labels.Matcher
 
 	if len(MustLabelValues(req.Context())) > 1 {
+		if r.regexMatch {
+			prometheusAPIError(w, "Only one label value allowed with regex match", http.StatusBadRequest)
```
simonpasquier, line 452:
```suggestion
			prometheusAPIError(w, "Only one label value allowed with regex match", http.StatusBadRequest)
			return
```

README.md
simonpasquier: I wonder if we shouldn't anchor the regexp by default (like Prometheus does for relabelings)? It would at least prevent too-permissive configurations.
It'd also be good to require that the compiled regex doesn't match the empty string.

README.md
sathieu: Done.

main.go
sathieu: Done.

injectproxy/routes.go
sathieu: Done (good catch 🤝 ).

main.go
sathieu: Done.

README.md
sathieu: > I wonder if we shouldn't anchor the regexp by default (like Prometheus does for relabelings)? It would at least prevent too-permissive configurations.

Prometheus anchors already.

> It'd also be good to require that the compiled regex doesn't match the empty string.

Good idea.

sathieu:
@simonpasquier I've merged your suggestions, added tests for regexp matching empty strings, and added tests early when using `-label-value`.

main.go
```diff
@@ -125,6 +128,23 @@ func main() {
 	if errorOnReplace {
 		opts = append(opts, injectproxy.WithErrorOnReplace())
 	}
+	if regexMatch {
```
simonpasquier, line 131: (nit)
```suggestion

	if regexMatch {
```

injectproxy/silences_test.go
```diff
@@ -74,11 +75,47 @@ func TestListSilences(t *testing.T) {
 			labelv:  []string{"default", "something"},
 			expCode: http.StatusUnprocessableEntity,
 		},
+		{
+			// Regex match
```
simonpasquier, line 79: hmm like with multiple label values, the silences API semantics don't allow to match multiple values.

https://github.com/prometheus-community/prom-label-proxy/blob/0664aa864a4e8aa5858d2772a0576cb3e385bb1f/injectproxy/routes.go#L305-L317

I'd recommend to update `assertSingleLabelValue()` and reject API calls when regex=true.

injectproxy/routes_test.go
```diff
@@ -1121,6 +1122,63 @@ func TestQuery(t *testing.T) {
...
+			expResponse:  okResponse,
+		},
+		{
+			name:       `HTTP header with invalid regexp with same regexp in query`,
+			headers:    http.Header{"namespace": []string{"tenant1-(.*"}},
+			headerName: "namespace",
+			regexMatch: true,
+			promQuery:  `up{instance="localhost:9090"} + foo{namespace="tenant1-.*"}`,
+			expCode:    http.StatusBadRequest,
+			// {"error":"error parsing regexp: missing closing ): `tenant1-(.*`","errorType":"prom-label-proxy","status":"error"}
```
simonpasquier, line 1162:
```suggestion
```

injectproxy/routes_test.go
```diff
@@ -1121,6 +1122,63 @@ func TestQuery(t *testing.T) {
...
+		},
+		{
+			name:       `Multiple regexp HTTP headers is invalid`,
+			headers:    http.Header{"namespace": []string{"tenant1", "tenant2"}},
+			headerName: "namespace",
+			regexMatch: true,
+			promQuery:  `up{instance="localhost:9090"} + foo{namespace="tenant1-.*"}`,
+			expCode:    http.StatusBadRequest,
+			// {"error":"Only one label value allowed with regex match","errorType":"prom-label-proxy","status":"error"}
```
simonpasquier, line 1171:
```suggestion
```

injectproxy/routes_test.go
```diff
@@ -1121,6 +1122,63 @@ func TestQuery(t *testing.T) {
...
+		},
+		{
+			name:       `Regex should not match empty string`,
+			headers:    http.Header{"namespace": []string{".*"}},
+			headerName: "namespace",
+			regexMatch: true,
+			promQuery:  `up{instance="localhost:9090"} + foo{namespace="tenant1-.*"}`,
+			expCode:    http.StatusBadRequest,
+			// {"error":"Regex should not match empty string","errorType":"prom-label-proxy","status":"error"}
```
simonpasquier, line 1180:
```suggestion
```

main.go
sathieu: Done.

injectproxy/silences_test.go
sathieu: `501 Not implemented` is now returned.

injectproxy/routes_test.go
sathieu: Done.

injectproxy/routes_test.go
sathieu: Done.

sathieu:
@simonpasquier Thanks for your review. I've merged your suggestions, and silence API is now returning `501 Not Implemented` when regex match is enabled. I'll try to add support for multiple values in silences in a future PR.

Back to you 🏸 !

injectproxy/routes_test.go
sathieu: Done.

injectproxy/routes.go
```diff
@@ -301,22 +311,41 @@ func NewRoutes(upstream *url.URL, label string, extractLabeler ExtractLabeler, o
 		)
 	}
 
-	errs.Add(
-		// Reject multi label values with assertSingleLabelValue() because the
-		// semantics of the Silences API don't support multi-label matchers.
-		mux.Handle("/api/v2/silences", r.el.ExtractLabel(
-			enforceMethods(
-				assertSingleLabelValue(r.silences),
+	if opt.regexMatch {
```
simonpasquier, line 314: (suggestion) to avoid duplicating the mux.Handle() calls, we might do something like this:

```go
errs.Add(
    mux.Handle("/api/v2/silences", r.errorIfRegexpMatch(
        r.el.ExtractLabel(...

[...]

 func (r *routes) errorIfRegexpMatch(next http.HandlerFunc) http.HandlerFunc {
    return func(w http.ResponseWriter, req *http.Request) {
        if r.regexMatch {
            prometheusAPIError(w, "support for regex match not implemented", http.StatusNotImplemented)
            return
        }

        return next(w, req)
    }
}

```

simonpasquier (commented):
one last comment...

injectproxy/routes.go
sathieu: Done @simonpasquier. Back to you ⚾ !

README.md
```diff
@@ -144,6 +144,30 @@ prom-label-proxy \
...
+```
+prom-label-proxy \
+   -label-value '^foo-.+$' \
+   -label namespace \
+   -upstream http://demo.do.prometheus.io:9090 \
+   -insecure-listen-address 127.0.0.1:8080 \
+   -regex-match
+```
+
+> :warning: The above feature is experimental. Be careful when using this option, it can expose sensitive metrics by accident.
```
simonpasquier, line 158: (nit) just to clarify that the onus is on the person configuring prom-label-proxy (e.g. it's not like prom-label-proxy would disclose data by itself).
```suggestion
> :warning: The above feature is experimental. Be careful when using this option, it may expose sensitive metrics if you use a too permissive expression.
```

simonpasquier (commented):
Thanks! :+1: for me.
@squat do you have any concern with this addition? From my side, the warning in the README.md is clear enough that users know about the potential issues if they mess up with their regexp.

simonpasquier (approved):

PR #184, opened 2024-01-26
chore: dont rely on slice labels

MichaHoffmann:
I wanted to proceed a bit on trying to convert thanos to build with stringlabels and prom-label-proxy is used as a library there. This would fix this build error for me!

### DIFF
--- a/injectproxy/rules.go
+++ b/injectproxy/rules.go
@@ -207,11 +207,8 @@ func (r *routes) filterRules(lvalues []string, resp *apiResponse) (interface{}, 
 	for _, rg := range rgs.RuleGroups {
 		var rules []rule
 		for _, rule := range rg.Rules {
-			for _, lbl := range rule.Labels() {
-				if lbl.Name == r.label && slices.Contains(lvalues, lbl.Value) {
-					rules = append(rules, rule)
-					break
-				}
+			if lval := rule.Labels().Get(r.label); lval != "" && slices.Contains(lvalues, lval) {
+				rules = append(rules, rule)
 			}
 		}
 		if len(rules) > 0 {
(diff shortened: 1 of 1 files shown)


SuperQ (approved):

squat:
@MichaHoffmann can you give me a bit more context? How is this code affecting external behavior in the Thanos build?

SuperQ:
@squat There is the new "stringlabels" code in Prometheus itself. When switching to this implementation, accessing the slice directly is no longer allowed.

SuperQ:
Here's what you get if you build prom-label-proxy now:

```
$ go build --tags=stringlabels .
# github.com/prometheus-community/prom-label-proxy/injectproxy
injectproxy/rules.go:210:24: cannot range over rule.Labels() (value of type "github.com/prometheus/prometheus/model/labels".Labels)
injectproxy/rules.go:234:23: cannot range over alert.Labels (variable of type "github.com/prometheus/prometheus/model/labels".Labels)
```

With this change, the build is successful.

MichaHoffmann:
> @MichaHoffmann can you give me a bit more context? How is this code affecting external behavior in the Thanos build?

What @SuperQ said! Just adding context that I want to convert Thanos to be compatible with the stringlabels tag and there are more build failures, this was just one of them and it seemed easy enough to fix.

injectproxy/rules.go
```diff
@@ -207,11 +207,8 @@ func (r *routes) filterRules(lvalues []string, resp *apiResponse) (interface{},
 	for _, rg := range rgs.RuleGroups {
 		var rules []rule
 		for _, rule := range rg.Rules {
-			for _, lbl := range rule.Labels() {
-				if lbl.Name == r.label && slices.Contains(lvalues, lbl.Value) {
-					rules = append(rules, rule)
-					break
-				}
+			if rule.Labels().Has(r.label) && slices.Contains(lvalues, rule.Labels().Get(r.label)) {
```
bboreham, line 210: This would be more efficient calling just `Get` and checking for an non-empty result.

bboreham (commented):
I’m not a maintainer here but had a comment if efficiency is a concern.

injectproxy/rules.go
MichaHoffmann: That makes sense! Thank you

MichaHoffmann: changed it!

squat (approved):
Thanks for the context @SuperQ @MichaHoffmann 🏄‍♀️🏄‍♀️🏄‍♀️

bboreham:
And soon `-tags dedupelabels` - https://github.com/prometheus/prometheus/pull/12304

PR #192, opened 2024-02-26
feat(routes): allow multitenancy through different labels

mutefiRe:
In a larger system, maintaining consistent tenant identifiers across all metrics isn't always practical. For example, some metrics might use `tenant_id` while others rely on `page_id`, depending on the reporting service. Therefore, the proxy needs to decide which label to enforce for each metric.

Instead of using the label parameter in the `NewRoutes` function, we inject the label name in the same way as the label values. As a result, the `ExtractLabeler` handles determining which label name to enforce.


mutefiRe:
hi! Sorry this was not supposed to land in this repository.

PR #211, opened 2024-04-18
chore: use golden file for tests

simonpasquier:
Modifying existing tests for alerts and rules endpoints is very tedious as you need to do lots of manual edits. This change relies on golden files which can be automatically updated as the reponse payload evolves.


squat (approved):
Very cool

PR #223, opened 2024-06-06
Adds the optional ability to split the tenant request header by comma

csthomas1:
Currently, prom-label-proxy will accept multiple header values only if they correspond to separately-occurring header lines in the incoming HTTP request. Example request:

```
GET /api/v1/query?query=up HTTP/1.1
Host: localhost:9090
User-Agent: curl/7.68.0
X-Tenant: firsttenant
X-Tenant: secondtenant
Accept: */*
```
Attempting to pass multiple header values via list-style syntax as described in RFC 9110 (https://www.rfc-editor.org/rfc/rfc9110#name-field-lines-and-combined-fi), on the other hand, results in a single tenant value consisting of the entire comma-separated string. Example request:

```
GET /api/v1/query?query=up HTTP/1.1
Host: localhost:9090
User-Agent: curl/7.68.0
X-Tenant: firsttenant, secondtenant
Accept: */*
```
With the above sample request, the tenant label match expression becomes "firsttenant, secondtenant"

This is inconvenient for use with Custom Header configuration on Grafana Data Sources, which seem to be unable to pass more than one request header line for a given header name -- the last header value overwrites the first.

This pull request adds OPTIONAL support for parsing the tenant request header as a comma-separated list. To enable, add the parameter "-header-uses-list-syntax" to the prom-label-proxy's commandline.

When the list syntax option enabled, the sample request containing the "firsttenant, secondtenant" list given above ultimately produces the tenant label match expression "firsttenant|secondtenant".

### DIFF
--- a/injectproxy/routes.go
+++ b/injectproxy/routes.go
@@ -234,7 +234,8 @@ func (hff HTTPFormEnforcer) getLabelValues(r *http.Request) ([]string, error) {
 
 // HTTPHeaderEnforcer enforces a label value extracted from the HTTP headers.
 type HTTPHeaderEnforcer struct {
-	Name string
+	Name            string
+	ParseListSyntax bool
 }
 
 // ExtractLabel implements the ExtractLabeler interface.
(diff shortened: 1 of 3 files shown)


simonpasquier (approved):
Thanks!

PR #225, opened 2024-06-07
chore: refactor PromQL enforcer

simonpasquier:
This change renames the Enforcer struct to PromQLEnforcer to clarify its purpose. It also removes the error type casting in favor of error wrapping.

### DIFF
--- a/README.md
+++ b/README.md
@@ -144,7 +144,7 @@ prom-label-proxy \
 
 `prom-label-proxy` will enforce the `tenant=~"prometheus|alertmanager"` label selector in all requests.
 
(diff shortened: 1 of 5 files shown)


squat (approved):
I think this is cool! I'm sometimes hesitant about replying with the invalid (read: potentially malicious) query string in the http response, since that *could* lead to exploits, but this is escaped as JSON so it should be ok

simonpasquier:
You've got a good point regarding not returning the bad PromQL expression. I'll change this (it will also be consistent with the Prometheus API).

PR #226, opened 2024-06-07
fix: support regex for non-query endpoints

simonpasquier:
PR #171 implemented support for regex label values but only for the query endpoints. This change adds support for all other Prometheus API endpoints.

FYI @sathieu. I remember that we discussed about the issue with the Alertmanager API endpoints but nothing about the non-query Prometheus endpoints. Let me know if the change looks sane.

### DIFF
--- a/injectproxy/routes.go
+++ b/injectproxy/routes.go
@@ -378,6 +378,9 @@ func NewRoutes(upstream *url.URL, label string, extractLabeler ExtractLabeler, o
 		"/api/v1/rules":  modifyAPIResponse(r.filterRules),
 		"/api/v1/alerts": modifyAPIResponse(r.filterAlerts),
 	}
+	//FIXME: when ModifyResponse returns an error, the default ErrorHandler is
+	//called which returns 502 Bad Gateway. It'd be more appropriate to treat
+	//the error and return 400 in case of bad input for instance.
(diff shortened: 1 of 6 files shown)


sathieu:
@simonpasquier Thanks for this PR. lgtm!

injectproxy/routes.go
```diff
@@ -378,6 +378,9 @@ func NewRoutes(upstream *url.URL, label string, extractLabeler ExtractLabeler, o
 		"/api/v1/rules":  modifyAPIResponse(r.filterRules),
 		"/api/v1/alerts": modifyAPIResponse(r.filterAlerts),
 	}
+	//FIXME: when ModifyResponse returns an error, the default ErrorHandler is
+	//called which returns 502 Bad Gateway. It'd be more appropriate to treat
+	//the error and return 400 in case of bad input for instance.
```
squat, line 383: 👍

injectproxy/routes_test.go
```diff
@@ -336,6 +338,33 @@ func TestMatch(t *testing.T) {
 			},
 			expBody: okResponse,
 		},
+		{
```
squat, line 341: Could you add a test case for the degenerate case where regex matches empty string?

squat (approved):
Great work @simonpasquier one comment but not blocking

PR #233, opened 2024-06-26
change: make `-error-on-replace` more cooperating

simonpasquier:
With this change, the proxy wouldn't necessary return an error if `-error-on-replace` is set and the incoming query has a label matcher overlapping with the enforced label.

For instance when the enforced label is `namespace="foo"`, the following expressions aren't rejected anymore:
- `up{namespace!="bar"}`
- `up{namespace=~"foo|bar"}`
- `up{namespace!~"bar"}`

The following expressions are rejected as before:
- `up{namespace=""}`
- `up{namespace="bar"}`
- `up{namespace=~"bar|fred"}`
- `up{namespace!~"foo"}`

### DIFF
--- a/injectproxy/enforce.go
+++ b/injectproxy/enforce.go
@@ -152,28 +152,124 @@ func (ms PromQLEnforcer) EnforceNode(node parser.Node) error {
 	return nil
 }
 
-// EnforceMatchers appends the configured label matcher if not present.
-// If the label matcher that is to be injected is present (by labelname) but
-// different (either by match type or value) the behavior depends on the
-// errorOnReplace variable and the enforced matcher(s):
-// * if errorOnReplace is true, an error is returned,
-// * if errorOnReplace is false and the label matcher type is '=', the existing matcher is silently replaced.
-// * otherwise the existing matcher is preserved.
+// EnforceMatchers appends the enforced label matcher(s) to the list of matchers
+// if not already present.
+//
+// If the label matcher that is to be injected is present (by labelname), the
+// behavior depends on the errorOnReplace variable and the enforced matcher(s):
+// * If errorOnReplace is false
+//   - And the label matcher type is '=', the existing matcher is silently
+//     discarded whatever is the original value.
+//   - Otherwise the existing matcher is preserved.
+//
+// * if errorOnReplace is true
+//   - And the label matcher and the enforced matcher are disjoint, the function returns an error.
+//   - Otherwise the existing matcher is preserved.
 func (ms PromQLEnforcer) EnforceMatchers(targets []*labels.Matcher) ([]*labels.Matcher, error) {
 	var res []*labels.Matcher
 
 	for _, target := range targets {
-		if matcher, ok := ms.labelMatchers[target.Name]; ok {
-			// matcher.String() returns something like "labelfoo=value"
-			if ms.errorOnReplace && matcher.String() != target.String() {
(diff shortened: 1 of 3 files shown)


simonpasquier:
@squat kindly asking for a review here. I added as many unit tests as possible to cover the different possibilities but I understand that the change isn't so easy to audit.

injectproxy/enforce.go
```diff
@@ -152,30 +152,126 @@ func (ms PromQLEnforcer) EnforceNode(node parser.Node) error {
...
+		}
+
+		if ms.errorOnReplace {
+			var ok bool
+
+			// Ensure that the expression's matcher combined with the
+			// enforced matchers can return some result. If the combined
+			// matchers return no result, the function returns an error.
+			//
+			// For instance, when te enforced matcher is 'tenant="bar"':
```
squat, line 185:
```suggestion
			// For instance, when the enforced matcher is 'tenant="bar"':
```

injectproxy/enforce_test.go
```diff
@@ -357,3 +366,533 @@ func TestEnforce(t *testing.T) {
...
+					`job!~"foo|bar"`,
+					`up{job!="foo",job!~"foo|bar"}`,
+					false,
+				},
+			},
+		},
+
+		// Regexp matcher enforcer.
+		{
+			enforcedMatcher: mustNewMatcher(labels.MatchRegexp, "job", "foo|bar"),
```
squat, line 626: there's some weird edge cases here that i'm not confident in because it depends so much on how the proxy's plumbing handles duplicate values. It's somewhat degenerate but I think it's possible but for an enforced matcher like `foo|foo` to be created. This would be disjoint with selectors like `job="bar"` and we *should* return errors in those cases since we are replacing things, right?

injectproxy/enforce_test.go
```diff
@@ -357,3 +366,533 @@ func TestEnforce(t *testing.T) {
...
+					`up{job="bar",job=~"foo|bar"}`,
+					false,
+				},
+				{
+					`job="fred"`,
+					``,
+					true,
+				},
+
+				// Not equal label selector in the expression.
```
squat, line 657: there's some weird edge cases here that i'm not confident in because it depends so much on how the proxy's plumbing handles duplicate values. It's somewhat degenerate but I think it's possible but for an enforced matcher like `foo|foo` to be created. This would be disjoint with selectors like `job!="foo"` and we *should* return errors in those cases since we are replacing things, right?
In other words, just because the matcher `labels.MatchRegexp` and the selector is `labels.MatchNotEqual` doesn't mean `ok`. WDYT?

injectproxy/enforce_test.go
```diff
@@ -357,3 +366,533 @@ func TestEnforce(t *testing.T) {
...
+					`up{job=~"foo",job=~"foo|bar"}`,
+					false,
+				},
+				{
+					`job=~"bar"`,
+					`up{job=~"bar",job=~"foo|bar"}`,
+					false,
+				},
+				{
+					`job=~"fred"`,
```
squat, line 696: these are disjoint, no? so why don't we error here?

squat (commented):
Thanks for the verbosity @simonpasquier. Overall I think it's pretty easy to follow. I have a couple of questions that have to do with the correctness of some weird cases, but otherwise it looks good I think.

injectproxy/enforce_test.go
simonpasquier: good catch!

PR #237, opened 2024-07-31
feat: allow to return rules with matching active alerts

simonpasquier:
This change adds a `-rules-with-active-alerts` CLI argument to return rules with active alerts matching the enforced label. The use case is for an admin to write general-purpose alerting rules which individual users can still retrieve from the API.

To avoid surprises ("why do I see this alert now?"), the feature needs to be explicitly enabled.

### DIFF
--- a/injectproxy/rules.go
+++ b/injectproxy/rules.go
@@ -174,7 +174,7 @@ var errModifyResponseFailed = errors.New("failed to process the API response")
 // modifyAPIResponse unwraps the Prometheus API response, passes the enforced
 // label value and the response to the given function and finally replaces the
 // result in the response.
-func modifyAPIResponse(f func([]string, *apiResponse) (interface{}, error)) func(*http.Response) error {
+func modifyAPIResponse(f func([]string, *http.Request, *apiResponse) (interface{}, error)) func(*http.Response) error {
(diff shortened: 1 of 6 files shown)


injectproxy/rules.go
```diff
@@ -223,9 +223,51 @@ func (r *routes) filterRules(lvalues []string, resp *apiResponse) (interface{},
...
+			}
+
+			var ar *alertingRule
+			for i := range rgr.alertingRule.Alerts {
+				if lval := rgr.alertingRule.Alerts[i].Labels.Get(r.label); lval == "" || !m.Matches(lval) {
+					continue
+				}
+
+				if ar == nil {
+					ar = &alertingRule{
```
jkroepke, line 243: I have one question here: Could you explain to me, why we have to copy the struct instead using 

```
ar = rgr.alertingRule 
```

I thought, after go 1.22, the loopvar issue is gone?

squat (approved):
This looks pretty good to me. I'm not too familiar with the motivation behind the feature request but it seems reasonable.

injectproxy/rules.go
jkroepke: Thanks for the clarification

PR #269, opened 2025-02-06
fix: use upstream path in modify response

seanbanko:
Some open-source projects expose a Prometheus-compatible API at at a subpath, e.g. `/prometheus`. We should use the `upstream.Path` configured when matching the `resp.Request.URL.Path` in `ModifyResponse` so that we can handle the case of a non-root upstream path when proxying `/api/v1/rules` and `/api/v1/alerts`.


squat:
Hi @seanbanko can you add a concrete test case? Otherwise I'm sure this will regress

seanbanko:
@squat just bumping this. Is that test case is sufficient?

seanbanko:
Just taking a look at recent commits, @simonpasquier are you able to review?

injectproxy/routes.go
```diff
@@ -389,8 +389,8 @@ func NewRoutes(upstream *url.URL, label string, extractLabeler ExtractLabeler, o
 
 	r.mux = mux
 	r.modifiers = map[string]func(*http.Response) error{
-		"/api/v1/rules":  modifyAPIResponse(r.filterRules),
-		"/api/v1/alerts": modifyAPIResponse(r.filterAlerts),
+		upstream.Path + "/api/v1/rules":  modifyAPIResponse(r.filterRules),
```
simonpasquier, line 392: Should we use https://pkg.go.dev/net/url#JoinPath instead?

It might be the right time to use a proper struct for modifies instead of an opaque map[string]func.

seanbanko: Updated to use `url.JoinPath`.

I agree the map for `modifies` isn't ideal. Do you think it's better to have a separate PR for the refactor to keep scope limited? Glad to open a follow-up.

squat:
I just re-ran the tests and they are still failing.
ping @seanbanko

seanbanko:
@squat @SuperQ fixed, apologies for the delay.

seanbanko:
Hi @simonpasquier , just following up on this to see if the submission is still being considered.

PR #283, opened 2025-05-12
chore: cut v0.11.1

simonpasquier:
Let's have a patch release to avoid people complaining about potential CVEs.

cc @squat


squat (approved):
Great idea @simonpasquier

PR #295, opened 2025-07-29
feat: allow to filter rules by label matchers

### LINKED ISSUE #236: Make use of label matcher support in Rules API
https://github.com/prometheus/prometheus/pull/10194 added label matcher support to the Rules API. prom-label-proxy could inject label matchers instead of unmarshaling/filtering/marshaling back the server's response (as it does today). We need to preserve the current implementation (since it's dependent on the server's version) and add a CLI flag to enable the new mode.

simonpasquier:
Closes #236


squat (approved):
Nice work @simonpasquier
Good job making the warning really clear in the flagset definition too. This could cause leaks otherwise

PR #301, opened 2025-08-29
chore: apply changes from Go modernize

simonpasquier:
https://pkg.go.dev/golang.org/x/tools/gopls/internal/analysis/modernize


squat (approved):
Not bad!

PR #349, opened 2026-04-08
Updates Prometheus and refactors to accommodate parser options breaking change

philipgough:
As per title!

I've tried to do this is a non-breaking change here but perhaps the alternative would be better.


simonpasquier (commented):
thanks! For `injectproxy/enforce.go`, I'd rather have the options passed to NewPromQLEnforcer() rather than a secondary call. If we want to limit disruption to consumers, we can create a `NewPromQLEnforcerWithOptions()` function.

injectproxy/routes.go
```diff
@@ -699,16 +716,17 @@ func (r *routes) matcher(w http.ResponseWriter, req *http.Request) {
 	r.handler.ServeHTTP(w, req)
 }
 
-func injectMatcher(q url.Values, matcher *labels.Matcher) error {
+func injectMatcher(q url.Values, matcher *labels.Matcher, parserOpts parser.Options) error {
```
simonpasquier: (nit) any reason not to make it a pointer receiver function?

```suggestion
func (r *routes) injectMatcher(q url.Values, matcher *labels.Matcher) error {
```

simonpasquier (commented):
one small nit, otherwise lgtm. I also realize that we have no unit test exercising the parser options but it can be added as a follow-up.

simonpasquier (approved):

PR #357, opened 2026-04-17
Update PromCI

SuperQ:
Update to latest PromCI.
* Cleanup CircleCI config.


.github/dependabot.yml
```diff
@@ -1,15 +1,17 @@
...
     schedule:
-      interval: monthly
-    open-pull-requests-limit: 10
-  - package-ecosystem: github-actions
-    directory: /
+      interval: "monthly"
+    groups:
+      aws:
+        patterns:
+          - "github.com/aws/*"
```
squat: why only updating AWS deps monthly?

SuperQ: Oh, I didn't meant to copy that pattern over.

This doesn't update only the pattern monthly, it just groups the updates together.

squat (approved):

PR #361, opened 2026-04-22
feat: Add path prefix upstream

### LINKED ISSUE #354: Support path prefix
We could not get prom-label-proxy to work with our external Mimir installation. All queries returned 404.

We believe the error is caused because Mimir exposes its Prometheus API under `/prometheus` and prom-label-proxy does not seem to support paths in its upstream.

We tried setting `--upstream` to `https://mimir.our.org.domain/prometheus` and `https://mimir.our.org.domain`. Accessing `http://prom-label-proxy:8080/prometheus/api/v1/query?query=up` and `http://prom-label-proxy:8080/api/v1/query?query=up` returns 404 in all cases.

Accessing Mimir directly via `https://mimir.our.org.domain/prometheus/api/v1/query?query=up` returns the metrics, as expected.

Also, adding some kind of debug logs would really help. #336

yuvaldekel:
This PR enables proxying to an upstream service with a path prefix. It introduces the `--upstream-path-prefix` flag, which appends a specified path to all forwarded requests.

Resolves #354


simonpasquier:
I'm not sure that we need this. Using `-upstream=http://example.com/some-prefix` should work IMHO.

simonpasquier:
Commented on #354 but my position is that we need first to understand why it doesn't work today. Implementing #336 would help.

SuperQ:
This needs a rebase

simonpasquier:
@yuvaldekel https://github.com/prometheus-community/prom-label-proxy/pull/361#issuecomment-4386139804 still stands

PR #377, opened 2026-06-05
Support client cert/key and server name for upstream TLS

### LINKED ISSUE #136: bypass https upstream tls
when using https endpoint, I am getting `http: proxy error: x509: certificate signed by unknown authority` is there a way to allow insecure upstream

ulascansenturk:
## Support client cert/key and server name for upstream TLS

Fixes #136.

The proxy already supported a custom CA (`-upstream-ca-cert`) and skipping verification (`-insecure-skip-verify`) for the upstream, but not the rest of the usual TLS client settings. As noted in #136, client certificate/key (mutual TLS) and an explicit TLS server name were missing.

This adds:

- `-upstream-client-cert` and `-upstream-client-key`: present a client certificate for mutual TLS with the upstream. Both must be set together.
- `-upstream-server-name`: the server name used to verify the upstream's certificate (also used for SNI), useful when the upstream URL host does not match the certificate.

Matching `injectproxy` options `WithUpstreamClientCert(certFile, keyFile)` and `WithUpstreamServerName(name)` are added alongside the existing `WithUpstreamCaCert` / `WithInsecureSkipVerify`.

### Tests

There were no existing tests for the upstream TLS configuration, so this adds `injectproxy/routes_tls_test.go`:

- verifies the client certificate is loaded and the server name is set on the proxy transport's `tls.Config`;
- verifies the error cases: only one of cert/key provided, and unreadable cert/key files.

`go test ./...`, `go vet ./...` and `gofmt` are clean.

### DIFF
--- a/injectproxy/routes.go
+++ b/injectproxy/routes.go
@@ -507,7 +507,7 @@ func NewRoutes(upstream *url.URL, label string, extractLabeler ExtractLabeler, o
 
 		cert, err := tls.LoadX509KeyPair(opt.upstreamClientCertFile, opt.upstreamClientKeyFile)
 		if err != nil {
-			return nil, fmt.Errorf("failed to load client certificate/key pair: %v", err)
+			return nil, fmt.Errorf("failed to load client certificate/key pair: %w", err)
 		}
 
 		transport.TLSClientConfig.Certificates = []tls.Certificate{cert}
--- a/injectproxy/routes.go
+++ b/injectproxy/routes.go
@@ -61,6 +61,9 @@ type routes struct {
 
 type options struct {
 	upstreamCaCert           string
+	upstreamClientCertFile   string
+	upstreamClientKeyFile    string
+	upstreamServerName       string
 	enableLabelAPIs          bool
 	passthroughPaths         []string
 	insecureSkipVerify       bool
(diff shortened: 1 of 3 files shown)


injectproxy/routes.go
```diff
@@ -479,6 +500,19 @@ func NewRoutes(upstream *url.URL, label string, extractLabeler ExtractLabeler, o
...
 	}
 
+	if opt.upstreamClientCertFile != "" || opt.upstreamClientKeyFile != "" {
+		if opt.upstreamClientCertFile == "" || opt.upstreamClientKeyFile == "" {
+			return nil, fmt.Errorf("both client certificate and key files must be provided")
+		}
+
+		cert, err := tls.LoadX509KeyPair(opt.upstreamClientCertFile, opt.upstreamClientKeyFile)
+		if err != nil {
+			return nil, fmt.Errorf("failed to load client certificate/key pair: %v", err)
```
simonpasquier: (nit)
```suggestion
			return nil, fmt.Errorf("failed to load client certificate/key pair: %w", err)
```

ulascansenturk: Done, switched to %w.

simonpasquier (approved):
thanks!

PR #383, opened 2026-07-01
feat: Add --rewrite-host-header-to option for upstream requests

### LINKED ISSUE #135: [feature request] add support for host header rewrite

dhanushm7:
This adds a `--rewrite-host-header-to` CLI flag that allows overriding the Host header sent to the upstream URL when proxying requests.

When prom-label-proxy is deployed in front of an upstream Prometheus endpoint accessed via an ingress controller with host-based routing, the proxy forwards the original client's Host header. This causes the ingress to fail routing because it cannot match the expected hostname.

The implementation switches from NewSingleHostReverseProxy to a ReverseProxy with a custom Rewrite function, which provides explicit control over the outbound Host header. When the flag is not set, behavior is unchanged.

Fixes #135.

### DIFF
--- a/injectproxy/routes.go
+++ b/injectproxy/routes.go
@@ -392,10 +392,9 @@ func NewRoutes(upstream *url.URL, label string, extractLabeler ExtractLabeler, o
 		Rewrite: func(r *httputil.ProxyRequest) {
 			r.SetURL(upstream)
 			r.SetXForwarded()
+			r.Out.Host = r.In.Host
 			if opt.rewriteHostHeader != "" {
 				r.Out.Host = opt.rewriteHostHeader
-			} else {
-				r.Out.Host = r.In.Host
 			}
 		},
 	}
--- a/injectproxy/routes.go
+++ b/injectproxy/routes.go
@@ -73,6 +73,7 @@ type options struct {
 	rulesWithActiveAlerts    bool
 	labelMatchersForRulesAPI bool
 	parserOptions            parser.Options
+	rewriteHostHeader        string
 }
 
 type Option interface {
@@ -193,6 +194,15 @@ func WithPromqlExtendedRangeSelectors() Option {
 func WithPromqlBinopFillModifiers() Option {
 	return optionFunc(func(o *options) {
 		o.parserOptions.EnableBinopFillModifiers = true
+	})
+}
+
+// WithRewriteHostHeader configures the proxy to rewrite the Host header
+// to the given value when proxying requests to the upstream. This is useful
+// when the upstream is behind an ingress that routes based on the Host header.
+func WithRewriteHostHeader(host string) Option {
+	return optionFunc(func(o *options) {
+		o.rewriteHostHeader = host
 	})
 }
 
@@ -378,7 +388,17 @@ func NewRoutes(upstream *url.URL, label string, extractLabeler ExtractLabeler, o
 		opt.registerer = prometheus.NewRegistry()
 	}
 
-	proxy := httputil.NewSingleHostReverseProxy(upstream)
+	proxy := &httputil.ReverseProxy{
+		Rewrite: func(r *httputil.ProxyRequest) {
+			r.SetURL(upstream)
+			r.SetXForwarded()
+			if opt.rewriteHostHeader != "" {
+				r.Out.Host = opt.rewriteHostHeader
+			} else {
+				r.Out.Host = r.In.Host
+			}
+		},
+	}
 
 	r := &routes{
 		upstream:              upstream,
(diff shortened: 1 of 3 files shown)


dhanushm7:
Hey @SuperQ / other maintainers, I'm following up on this when you have a moment. It fixes [#135 ](https://github.com/prometheus-community/prom-label-proxy/issues/135)by adding --rewrite-host-header-to, and is a clean rebase-with-tests version of the earlier stale #154. No pressure, just don't want it to fall off the radar. Thanks!

simonpasquier (commented):
I think that we need to [ProxyRequest.SetXForwarded()](https://pkg.go.dev/net/http/httputil@go1.26.5#ProxyRequest.SetXForwarded) call in the Rewrite() function to preserve the existing behavior.

injectproxy/routes.go
```diff
@@ -378,7 +388,16 @@ func NewRoutes(upstream *url.URL, label string, extractLabeler ExtractLabeler, o
...
 	}
 
-	proxy := httputil.NewSingleHostReverseProxy(upstream)
+	proxy := &httputil.ReverseProxy{
+		Rewrite: func(r *httputil.ProxyRequest) {
+			r.SetURL(upstream)
+			if opt.rewriteHostHeader != "" {
+				r.Out.Host = opt.rewriteHostHeader
+			} else {
+				r.Out.Host = r.In.Host
```
simonpasquier: it should already be the case
```suggestion
```

dhanushm7: Thanks for the review @simonpasquier! I've added `r.SetXForwarded()` as suggested.

I'd like to keep the `else` branch though. `NewSingleHostReverseProxy` used the `Director` API which never touches `req.Host` - only `req.URL.Host` (the TCP destination). So the client's original `Host:` header was preserved implicitly.

With the `Rewrite` API, `SetURL()` explicitly clears `r.Out.Host = ""`, causing Go to fall back to the upstream target's hostname as the `Host:` header. Without the `else` branch, users not using `--rewrite-host-header-to` would silently get a different `Host:` header than before - a behavioral regression. The Go docs for [SetURL](https://pkg.go.dev/net/http/httputil#ProxyRequest.SetURL) also explicitly show `r.Out.Host = r.In.Host` as the pattern to preserve the inbound Host.

Our test `TestRewriteHostHeader/original_host_preserved_when_option_is_not_set` confirms this - removing the `else` branch causes it to fail.

injectproxy/routes_host_test.go
```diff
@@ -0,0 +1,85 @@
...
+			name:        "original host preserved when option is not set",
+			rewriteHost: "",
+			wantHost:    "example.com",
+		},
+	} {
+		t.Run(tc.name, func(t *testing.T) {
+			var receivedHost, receivedXForwardedFor, receivedXForwardedHost, receivedXForwardedProto string
+			m := newMockUpstream(http.HandlerFunc(func(w http.ResponseWriter, req *http.Request) {
+				receivedHost = req.Host
+				receivedXForwardedFor = req.Header.Get("X-Forwarded-For")
```
simonpasquier: (suggestion) the test server can do the verification on headers and return a 400 status code + detailed message in case the request doesn't match the expectations.

injectproxy/routes.go
simonpasquier: You're correct! I'd still prefer the following form
```
		r.Out.Host = r.In.Host
		if opt.rewriteHostHeader != "" {
			r.Out.Host = opt.rewriteHostHeader
		}
```

injectproxy/routes_host_test.go
dhanushm7: Updated as per suggestion, thanks!

injectproxy/routes.go
dhanushm7: Updated as per suggestion, thanks!

simonpasquier:
LGTM but we need the commits to be signed

https://github.com/prometheus-community/prom-label-proxy/pull/383/checks?check_run_id=90559059991

simonpasquier (approved):

simonpasquier:
Thanks!

PR #400, opened 2026-07-22
feat: add support for multiple enforced labels

### LINKED ISSUE #190: Allow different label names injected by an ExtractLabeler
In a bigger system you don't always have the same tenant identifiers as a label in every single metric.
For example we could have a `tenant_id` or a `page_id` to be enforced because the service where the metric is reported would not know about the tenant_id, but about the `page_id`. The proxy should decide based on the metric which label has to be enforced.

We have a running prototype of this with modified vendor files where we do not use the label param in the `NewRoutes` function, but inject the label name the same way as the label values are injected.
So instead the `ExtractLabeler` has to take care of which label name must be enforced.

We don't need all the functionallity of the proxy, e.g. rules, silences, alerts, which seems harder to support - therefore I want to ask you if like the Idea or you interested in supporting this, then we can probably put more effort into polishing this for an upstream PR. Anyways we can provide a fork if it is ready.

lukasboettcher:
This PR adds support for enforcing multiple ExtractLabelers in a single prom-label-proxy instance.
The goal is to keep backwards compatibility for the CLI and API interface.

Multiple ExtractLabelers can be created by specifying `--label` and `--query-param` or `--header-name` arguments multiple times. The ordering is important to associate labels with their value sources. The `--label-value` flag is excluded from this feature due to the ambiguity of mapping static label values to multiple labels.

Disclaimer: this PR was created with the help of AI and manually refined

fixes: https://github.com/prometheus-community/prom-label-proxy/issues/190


examples/prometheus/compose.yaml
simonpasquier, line 1: please remove this

examples/prometheus/prometheus.yml
simonpasquier, line 1: same here

README.md
```diff
@@ -120,6 +120,22 @@ You can provide multiple values for the label using several HTTP headers:
...
+```
+prom-label-proxy \
+   -label tenant -header-name X-Tenant \
+   -label cluster -header-name X-Cluster \
+   -label environment -header-name X-Environment \
+   -upstream http://demo.do.prometheus.io:9090 \
+   -insecure-listen-address 127.0.0.1:8080
+```
+
+The same positional pairing applies when repeating `-query-param` instead of
```
simonpasquier: IMHO it isn't a very good user experience as it's easy to get wrong. Maybe we've reached the limit of what we can configure using CLI arguments and need to support file-based configuration?

simonpasquier (commented):
Thanks for starting this. 2 remarks
* Let's try to change only what's required. I get that help text could get better and be reworded but it should be part of a separate PR.
* I think that we should improve the proposed UX.

README.md
lukasboettcher: I thought about that and I agree that the positional pairing would be easy to get wrong. I didn't want to change any existing or introduce any new config params, but a new file based config would be the cleanest imo.

I can rework the PR to include a file based config.

SuperQ:
This needs a rebase

lukasboettcher:
rebased onto main, and added a file-based configuration for multiple enforced labels. the logic for the old cli arguments remains unchanged.

injectproxy/config.go
```diff
@@ -0,0 +1,173 @@
...
+	"go.yaml.in/yaml/v3"
+)
+
+// Config declares the labels enforced by the proxy.
+type Config struct {
+	Labels []LabelConfig `yaml:"labels"`
+}
+
+// LabelConfig declares a single enforced label and the source of its values.
+// Exactly one of Header, QueryParam or Values must be set.
```
simonpasquier, line 33: is there a use case where someone would need to mix the source of enforced label values? E.g. label `foo` comes from the query parameters and label `bar` comes from HTTP headers. Maybe static values are a legit case though.

injectproxy/config.go
```diff
@@ -0,0 +1,173 @@
...
+			return fmt.Errorf("labels[%d]: query parameter %q is used by more than one label", i, l.QueryParam)
+		}
+		seenQueryParams[l.QueryParam] = struct{}{}
+	}
+
+	return nil
+}
+
+func (l LabelConfig) validate() error {
+	if !model.UTF8Validation.IsValidLabelName(l.Name) {
```
simonpasquier: Please don't add changes which aren't strictly related to the original issue. It adds on the reviewer.

injectproxy/silences.go
```diff
@@ -45,14 +44,15 @@ func (r *routes) silences(w http.ResponseWriter, req *http.Request) {
 	}
 }
 
-// assertSingleLabelValue verifies that the proxy is configured to match only
-// one label value. If not, it will reply with "422 Unprocessable Content".
-func assertSingleLabelValue(next http.HandlerFunc) http.HandlerFunc {
+// assertSingleLabelValue verifies that each enforced label has only one value.
+// If not, it will reply with "422 Unprocessable Content".
+func (r *routes) assertSingleLabelValue(next http.HandlerFunc) http.HandlerFunc {
```
simonpasquier, line 53: why this change?

injectproxy/enforce.go
```diff
@@ -295,8 +296,10 @@ func (ms PromQLEnforcer) EnforceMatchers(targets []*labels.Matcher) ([]*labels.M
 		res = append(res, target)
 	}
 
-	for _, enforcedMatcher := range ms.labelMatchers {
-		res = append(res, enforcedMatcher)
+	// Iterate in a deterministic order so that the injected matchers are
+	// always rendered the same way.
```
simonpasquier: please make it a separate PR

injectproxy/routes.go
```diff
@@ -376,12 +388,47 @@ func (sle StaticLabelEnforcer) ExtractLabel(next http.HandlerFunc) http.Handler
 	})
 }
 
+// NewRoutes returns a proxy which enforces the configured labels on every
+// request forwarded to the upstream.
+//
+// The label and extractLabeler arguments declare the first enforced label. They
+// may be left empty and nil when the labels are declared with
+// WithEnforcedLabel() or WithConfig() instead.
```
simonpasquier: While I understand and appreciate the attention towards downstream users I feel that it complicates the API.
Can we have a separate constructor maybe and NewRoutes becoming a shim?

```
func NewRoutesWithLabelers(upstream *url.URL, labelers []LabelEnforcer, opts ...Option) {
...
}

NewRoutes(upstream *url.URL, labelers []LabelEnforcer, opts ...Option) {
   // build labelers list.
   ...
   return NewRoutesWithLabelers(...)
}
```

injectproxy/routes.go
```diff
@@ -589,21 +635,51 @@ type ctxKey int
...
+// extractLabels runs every configured extractor in turn. Each ExtractLabeler
+// reports its values through keyLabel (the single-label API kept for backward
+// compatibility) and the wrapper immediately re-keys them by label name.
+func (r *routes) extractLabels(next http.HandlerFunc) http.Handler {
+	var handler http.Handler = next
+
+	if len(r.labels) > 1 {
+		// keyLabel can only hold one label's values. Drop it so that
+		// MustLabelValue(s)() panics instead of silently returning the values
+		// of whichever label happened to be extracted last.
```
simonpasquier: I don't understand this part.

injectproxy/routes.go
```diff
@@ -714,51 +815,91 @@ func enforceQueryValues(e *PromQLEnforcer, v url.Values) (values string, noQuery
...
+	return matchers, nil
+}
+
+func (r *routes) newLabelMatcherFor(label string, vals ...string) (*labels.Matcher, error) {
 	if r.regexMatch {
 		if len(vals) != 1 {
 			return nil, errors.New("only one label value allowed with regex match")
 		}
 
-		re := vals[0]
```
simonpasquier, line 723: why remove the intermediate var?

injectproxy/config.go
lukasboettcher: in our case we have one prom-label-proxy instance for a sensitive label enforcement like X-Tenant and another chained instance with query parameters where the user can apply their own filtering in the grafana datasource for a datacenter label. it costs nothing and I couldn't think of a reason to restrict this in the config.

injectproxy/silences.go
lukasboettcher: `assertSingleLabelValue` needs access to all `r.labels` so making it a method made sense to me.

injectproxy/routes.go
lukasboettcher: no reason, just simplification. I reverted it

injectproxy/routes.go
lukasboettcher: i will add `NewRoutesWithLabelers` and removed `WithEnforcedLabel` and `WithConfig`

injectproxy/enforce.go
lukasboettcher: actually on second thought, I don't believe this is needed at all. this is sorted in the prometheus parser anyway

injectproxy/routes.go
lukasboettcher: that was a precaution proposed by claude. but should not be necessary, since all enforced labels have their own context key

